PR #354, opened 2024-01-26
feat: add filter on has case/has no case on decisions page

Pascal-Delange:
As discussed, I implemented this by allowing to filter by the different cases "is attached to a case" and "is not attached to a case".

Maybe not totally ideal compared to just allowing one of those two options (assuming only one of them really makes sense from a product perspective) where we could have worked more easily with a toggle in the filter, but let's keep it like this for now.

Capture d’écran 2024-01-29 à 11 11 17
Capture d’écran 2024-01-29 à 11 11 22
Capture d’écran 2024-01-29 à 11 11 29

The back-end API is a change compared to the current implementation, cf https://github.com/checkmarble/marble-backend/pull/462

### DIFF
--- a/packages/app-builder/public/locales/en/common.json
+++ b/packages/app-builder/public/locales/en/common.json
@@ -1,4 +1,6 @@
 {
+  "true": "True",
+  "false": "False",
   "auth.logout": "Log out",
   "search": "Search",
   "error_one": "{{count}} error",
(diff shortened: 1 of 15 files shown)


balzdur:
Refactor to make use of a single switch with opinionated default behaviour.

**Context:**
Filters are implemented with "auto save" on close. To make it work using a switch it means when we first open the modal, we need to force the boolean state to a value (in case it is undefined) and thus, closing the window without interacting with the switch will force the filter to the default value.
In order to remove, you need to use the "cross" inside the filter bar (cf video for more info on this).

Still I think this is preferable to the previous implementation: we no longer have a "presence of case" filter in the filter bar if we do not want to filter on this (= ot
output
her filters behaviours)

balzdur (approved):

PR #360, opened 2024-02-06
Thomas/mar 412 api keys page

balzdur:
Implement UI for API key management :
- list existing keys (desc + role) on settings
- delete existing keys
- create a new key
  - use auth session to store the new key as a cookie flash message (to ensure cleanning)
  - display once, with message to let you know you need to copy the value
  
output


packages/app-builder/public/locales/en/settings.json
```diff
@@ -23,6 +23,19 @@
...
+  "api_keys.role": "API key role",
+  "api_keys.role.admin": "ADMIN",
+  "api_keys.role.publisher": "PUBLISHER",
+  "api_keys.role.builder": "BUILDER",
+  "api_keys.role.viewer": "VIEWER",
+  "api_keys.role.api_client": "API_CLIENT",
+  "api_keys.role.unknown": "UNKNOWN",
+  "api_keys.create": "Create new API Key",
+  "api_keys.delete": "Delete API Key",
+  "api_keys.delete.content": "You are about to delete this API key. This action is irreversible, do you want to proceed?",
```
Pascal-Delange, line 38: same as in backend, I have a preference for "revoke" rather than "delete" (but that's just wording of course)

packages/app-builder/public/locales/en/settings.json
```diff
@@ -23,6 +23,19 @@
   "users.inbox_user_role.member_count_other": "MEMBER in {{count}} inboxes",
   "users.inbox_user_role.unknown": "unknown",
   "api_keys": "API Keys",
+  "api_keys.new_api_key": "New API Key",
+  "api_keys.copy_api_key": "Make sure to copy it now as you will not be able to see this again.",
```
Pascal-Delange, line 27: Plus "and store it in a secure place" or something along those lines ?

packages/app-builder/src/routes/_builder+/settings+/api-keys.tsx
```diff
@@ -1,3 +1,135 @@
...
+    {
+      headers,
+    },
+  );
+}
+
+const columnHelper = createColumnHelper<ApiKey>();
+
 export default function ApiKeys() {
-  return <div>ApiKeys</div>;
```
Pascal-Delange, line 2: I'm asking because I couldn't find it: did you put a check on the authenticated user role somewhere before displaying this ? Is it somewhere "higher up" in the file hierarchy ? Because effectively this should only be available for admins

Pascal-Delange (commented):
Some small comments in the PR, but mainly I have one question: do we really want (today) to officially allow the creation of API keys with roles other than "API client" for the customers ?
While I think the general principle is good and we'll get there eventually, it poses some general documentation + backwards compatibility questions (we currently don't consider our other APIs as public), as well as some tricky security questions (I'm unsure about org admin API keys as a concept).

packages/app-builder/src/routes/_builder+/settings+/api-keys.tsx
balzdur: It's "higher up" in the file, look at the `loader` :
```ts
const { apiKey, user } = await authService.isAuthenticated(request, {
    failureRedirect: getRoute('/sign-in'),
  });
  if (!isAdmin(user)) {
    return redirect(getRoute('/'));
  }
```

Pascal-Delange (approved):

PR #361, opened 2024-02-06
feat(scenario): update scenario name and description

balzdur:
Disclaimer: this is a fast implementation of name/description scenario edit. Since we do not have a scenario dedicated page (home or setting) there is no "good" place for this feature. That's why I took the decision to "hide" this behind the scenario name.

If you do not know this is clickable, nothing let you no it is.

This is done on purpose, to help Marble members easilly change scenario info w\ doing db manipulation but give the possibility to move this to a dedicated scenario page in the future (the goal is to not build strong UX memory/habits arround this)
output


Pascal-Delange (approved):
👍

PR #362, opened 2024-02-08
Thomas/handle UUID parameters

balzdur:
The goal of this PR is to solve the "raw decision uuid" in URL issue (more info in [this Slack discussion ref](https://checkmarble.slack.com/archives/C063SMVJ4BY/p1706622280182339))

**Solution:**
- introduce a specific zod schema to transform path param that could either be short or raw uuid
- create a reusable error handler to redirect to shorth uuid URL version

**Bonus:**
I took inspiration a lot of the kind of code structure we have in our backend, using Gin + Go. What I particularly appreciate is using "safe" parsing version (= explicitly return `val, err` couple like in Go): 
- handling the error responsibility is done in the integration code, so reading a loader/action make it explicit
- using `return` instead of `throw` trick (like in `auth` middleware for example) reading the code make it explicit when there is an expected early return (no more hidden "if this go wrong I throw and an error occure making the code early return")

In the end, this introduce explicit control flow in the code with a couterpart of more boilerplate (acceptable).

The long term plan : build other `handleError*` so we can remove the "throw when status code is considered wrong" behaviour of our client API (no need to try/catch since we can return wellformed `json(...)` response instead of built in `HHTPError` in `oazapfts`)

### DIFF
--- a/packages/app-builder/src/routes/_builder+/decisions+/$decisionId.tsx
+++ b/packages/app-builder/src/routes/_builder+/decisions+/$decisionId.tsx
@@ -13,8 +13,11 @@ import { ScorePanel } from '@app-builder/components/Decisions/Score';
 import { TriggerObjectDetail } from '@app-builder/components/Decisions/TriggerObjectDetail';
 import { isNotFoundHttpError } from '@app-builder/models';
 import { serverServices } from '@app-builder/services/init.server';
+import { handleParseParamError } from '@app-builder/utils/handle-errors';
+import { NOT_FOUND } from '@app-builder/utils/http-status-codes';
+import { parseParamsSafe } from '@app-builder/utils/input-validation';
 import { getRoute } from '@app-builder/utils/routes';
-import { fromParams } from '@app-builder/utils/short-uuid';
+import { shortUUIDSchema } from '@app-builder/utils/schema/shortUUIDSchema';
 import { json, type LoaderFunctionArgs } from '@remix-run/node';
 import {
   isRouteErrorResponse,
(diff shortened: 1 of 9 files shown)


packages/app-builder/src/utils/http-status-codes.ts
```diff
@@ -0,0 +1,361 @@
+/**
+ * The following is a list of HTTP status codes and their meanings.
+ *
+ * Copied from 'http-status-codes' package. This is a copy of the package to avoid a dependency on the package and allow tree shaking.
+ * @see https://www.npmjs.com/package/http-status-codes
```
Pascal-Delange, line 5: while we're at it, maybe include a line of notice that it's under MIT license

packages/app-builder/src/routes/_builder+/decisions+/$decisionId.tsx
```diff
@@ -38,15 +42,22 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
 
-  const decisionId = fromParams(params, 'decisionId');
   try {
     const decision = await apiClient.getDecision(decisionId);
 
     return json({ decision });
   } catch (error) {
     if (isNotFoundHttpError(error)) {
-      throw new Response(null, { status: 404, statusText: 'Not Found' });
+      throw json(null, NOT_FOUND);
```
Pascal-Delange, line 60:
```suggestion
      return json(null, NOT_FOUND);
```
unless I misunderstand?

Pascal-Delange (approved):

packages/app-builder/src/routes/_builder+/decisions+/$decisionId.tsx
Pascal-Delange: got it 👍  ça me va

PR #364, opened 2024-02-09
feat(analytics): add analytics page

balzdur:
Implementation of the new analytics page :
- handle permission
- retrieve data from the backend
- only use the `global_dashboard` for now
- display the embeded dashboard

Disclaimer: this is "untested" code, mostly because of a lack of knowledge on how metabase embeded works (= how do we inject/customize parameters) + `org_id` injected inside signed URL... locally based 🙃

Just in case, this is how it looks like
image


Pascal-Delange (approved):
all clear 👍

PR #366, opened 2024-02-15
WIP - stale

balzdur:
Trying to iterate other the existing API make me wonder if we do have the good representation.

Currently we define a specific "resource" to handle `scenario-publication` related usecases. More than just a renaming, I do face some consistency issues :
- Adding `*/publish`, `*/deactivate` ... seems weird: id we do have an array of "events" corresponding to those action, do we need separate endpoints for each type of action ?
- Validation of a scenario (= required to pass `publication`) is exposed by `/scenario-iterations/{scenarioIterationId}/validate`. If we plan to add a "is scenario activable" endpoint, where should I put the endpoint ? (it's not totally comparable since the validation is used for the AST, which basically is something we can interact with in the frontend, not like the activation requirements here)

### DIFF
--- a/packages/marble-api/scripts/openapi.yaml
+++ b/packages/marble-api/scripts/openapi.yaml
@@ -1583,91 +1583,170 @@ paths:
           $ref: '#/components/responses/403'
         '404':
           $ref: '#/components/responses/404'
-  /scenario-publications:
+  /scenario-life-cycle:
     get:
       tags:
-        - Scenario Publications
-      summary: List scenario publications
-      description: List scenario publications using provided filters
-      operationId: listScenarioPublications
+        - Scenario Life Cycle
+      summary: List scenario life cycle events
+      description: List scenario life cycle events corresponding to filters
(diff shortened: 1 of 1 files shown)


Pascal-Delange (commented):
Following what we said this morning: let's first split the publication endpoint while keeping the "publication" part naming identical, and focus first on the "data preparation part".
I'm starting work on the endpoint splitting (hard to advance much on Fri morning with all the meetings), can you propose an API contract for the index creation side of things ?

balzdur:
I can, but indeed it's more or less what the `POST /scenario-life-cycle/activate` is meant to.
The only endpoint I do not include in this PR is the one that check "activeness".

Still, I will push a new proposition with minimal API changes, without regarding too much on consistency. The goal is to stay pragmatic, by introducing the minimum changes into the API since we do not have clear minds arround it.

What we really need is a backend implementation of usecases + frontend integration (should be seamless for our first version) so we can move forward on this.

PR #367, opened 2024-02-16
WIP

balzdur:
Introduce two endpoints on the `/scenario-publications/` resrouce :
- GET: check the current status of the preparation required for publication (= index creation, validation...)
- POST: start preparation async work

### DIFF
--- a/packages/marble-api/scripts/openapi.yaml
+++ b/packages/marble-api/scripts/openapi.yaml
@@ -1657,6 +1657,78 @@ paths:
           $ref: '#/components/responses/403'
         '404':
           $ref: '#/components/responses/404'
+  /scenario-publications/preparation:
+    get:
+      tags:
+        - Scenario Publications
+      summary: Get the preparation status of a scenario publication
+      operationId: getScenarioPublicationPreparationStatus
+      security:
+        - bearerAuth: []
+      requestBody:
+        description: 'Describe the scenario publication action'
+        content:
+          application/json:
+            schema:
+              type: object
+              required:
+                - scenario_iteration_id
+              properties:
+                scenario_iteration_id:
+                  type: string
+                  format: uuid
+        required: true
+      responses:
+        '200':
+          description: The preparation status of the scenario publication
+          content:
+            application/json:
+              schema:
+                type: object
+                required:
+                  - preparation_status
+                properties:
+                  preparation_status:
+                    type: string
+                    enum:
+                      - 'PENDING'
+                      - 'IN_PROGRESS'
+                      - 'DONE'
+        '401':
+          $ref: '#/components/responses/401'
+        '403':
+          $ref: '#/components/responses/403'
+        '404':
+          $ref: '#/components/responses/404'
+    post:
+      tags:
+        - Scenario Publications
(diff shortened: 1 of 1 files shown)


packages/marble-api/scripts/openapi.yaml
```diff
@@ -1657,6 +1657,78 @@ paths:
...
+              required:
+                - scenario_iteration_id
+              properties:
+                scenario_iteration_id:
+                  type: string
+                  format: uuid
+        required: true
+      responses:
+        '204':
+          description: The preparation of the scenario publication has been started
```
balzdur, line 1725: Non bloquant call (we can decide to await on the creation async work if needed but IIUC this is not what we want)

packages/marble-api/scripts/openapi.yaml
```diff
@@ -1657,6 +1657,78 @@ paths:
...
+                type: object
+                required:
+                  - preparation_status
+                properties:
+                  preparation_status:
+                    type: string
+                    enum:
+                      - 'PENDING'
+                      - 'IN_PROGRESS'
+                      - 'DONE'
```
balzdur, line 1696: This is meant to change depending of what we really need but you have the idea: expose information based on current preparation status.
The granularity depends on :
- what we can really get from backend (ex: index creation status)
- what we really want to display to the user

Pascal-Delange: all right,
I'd suggest for now the following values:
```
- 'PREPARATION_REQUIRED'
- 'PREPARATION_IN_PROGRESS'
- 'READY_TO_ACTIVATE'
```
While keeping in mind the fact that we won't be able to tell for which iteration the (a) preparation is in progress

packages/marble-api/scripts/openapi.yaml
Pascal-Delange: Well the go server will reply anyway (while doing work in the background).
Later we could image to do something advanced where we wait for up to N seconds and then return with a "work is continuing in the background) perhaps.

packages/marble-api/scripts/openapi.yaml
balzdur: I would suggest to distinguish things then :
```ts
type PreparationStatus = 'PREPARATION_REQUIRED' | 'PREPARATION_IN_PROGRESS' | 'READY_TO_ACTIVATE';

type PreparationServiceStatus = 'AVAILABLE' | 'OCCUPIED' // a temporary "global" state we need to use until a queue system or equivalent is introduced
```
NB: I know we can't return `'PREPARATION_IN_PROGRESS'` for now but it's a future possibility. It's just to illustrate we may need more than just a "boolean" state for that.

packages/marble-api/scripts/openapi.yaml
balzdur: We can change for a `202` status code, but it's not really important.

packages/marble-api/scripts/openapi.yaml
Pascal-Delange: agreed

PR #370, opened 2024-02-19
Thomas/introduce commit draft

balzdur:
First step of the scenario iteration life cycle rework: 
- distinguish between activate and commit
- refactor deployment code

NB: add a "commit" Icon (need to be validated) + suppression of success modal (really anoying to handle and less usefull with the current behaviour)

We can still dispatch toast messages to give more feedback


Pascal-Delange (approved):
👍

PR #371, opened 2024-02-20
Thomas/introduce prepare version

balzdur:
In this PR, introduction of preparation state :
- check if the version need preparation
- handle some errors to give contextual toast error messages to the user

**Bonus:** new convention to handle toast message: now we translate the error messages directly in the backend. It's easier to manipulate, and most of all it allows :
- group translation errors definition close to where it's needed
- allow the use of translation keys with parameters (since we do translate the message in the server, we can format it using some structured data)


packages/app-builder/src/models/toast-session.ts
```diff
@@ -16,10 +21,29 @@ export const toastMessageScema = z.object({
...
 
+/**
+ * @param message the translated message
+ */
+const newToastMessageSchema = z.object({
+  type: z.enum(['success', 'error', 'loading', 'custom']),
+  message: z.string(),
+});
+
+export const toastMessageScema = z.union([
```
Pascal-Delange, line 35:
```suggestion
export const toastMessageSchema = z.union([
```

Pascal-Delange (approved):
👍 
Let's just include the proper error code handling and we're good ! (or we can just go and merge it by checking the error code string without decoding the full json, should work as well for now - as you want)

PR #373, opened 2024-02-21
change the control flow for the activation of a scenario version

balzdur:
Goal: improve future proof of the code


packages/app-builder/src/routes/_builder+/scenarios+/$scenarioId+/i+/$iterationId+/_edit-view+/_layout.tsx
```diff
@@ -238,17 +238,17 @@ function DeploymentActions({
         <CommitScenarioDraft scenarioId={scenario.id} iteration={iteration} />
       );
     case 'version':
-      if (iteration.status === 'required') {
+      if (iteration.status === 'ready_to_activate') {
```
balzdur, line 241: Better to display `activate` only when the status is `ready_to_activate`.

The old code is totally equivalent _for now_ but the day we push a "pending" state it won't.

I prefer this control flow structure

Pascal-Delange (approved):

PR #375, opened 2024-02-21
segment write key for open source users

Pascal-Delange:
New api key created for opensource

### DIFF
--- a/packages/app-builder/.env.example
+++ b/packages/app-builder/.env.example
@@ -26,7 +26,7 @@ FIREBASE_APP_ID=dummy
 
 
 # we use it for analytics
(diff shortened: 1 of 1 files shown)


packages/app-builder/.env.example
```diff
@@ -26,7 +26,7 @@ FIREBASE_APP_ID=dummy
 
 
 # we use it for analytics
-SEGMENT_WRITE_KEY=hC8qrY2OLhUpl1Xycw523tbuClxlQR6u
```
Pascal-Delange, line 29: this is the dev env key, which I presume we'll share internally only

balzdur (approved):

PR #376, opened 2024-02-21
refactor(validation): use a single endpoint for validation

balzdur:
In this PR :
- remove the GET validation endpoint from the API
- refactor validation related repository + `validate-with-given-trigger-or-rule`: the main idea is to expose higher level resources to the frontend
   - get trigger specific validation
   - get rule specific validation

In the future, we may want to only validate a portion of the ast or equivalent, this interface makes it easier to change (and understand what's really going on)

### DIFF
--- a/packages/app-builder/src/repositories/ScenarioRepository.ts
+++ b/packages/app-builder/src/repositories/ScenarioRepository.ts
@@ -18,6 +18,7 @@ import {
   type ScenarioIterationSummary,
   type ScenarioPublicationStatus,
 } from '@app-builder/models/scenario';
+import { findRuleValidation } from '@app-builder/services/validation/scenario-validation';
 import { type Scenario } from 'marble-api';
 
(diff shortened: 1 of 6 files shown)


packages/app-builder/src/repositories/ScenarioRepository.ts
```diff
@@ -45,10 +46,15 @@ export interface ScenarioRepository {
     scenarioId: string;
   }): Promise<ScenarioIterationSummary[]>;
   validate(args: { iterationId: string }): Promise<ScenarioValidation>;
-  validateScenarioIterationTrigger(args: {
+  validateTrigger(args: {
```
Pascal-Delange, line 49: pretty nice, that's a good usecase of having an abtraction over the REST backend API in the front repo 👍

Pascal-Delange (approved):

PR #377, opened 2024-02-21
fix(firebase): add logout method to firebase service

balzdur:
This is an attempt to fix the "double login" issue (first one result in an error, second one works).

### Context
From time to time, users experiment a "first login attempt failure". 

It seems to be related to a Firebase error due to a lack of permission on the token. It may be due to the current `logout` process which clear the user session but not the firebase client side data.

In order to ensure a Firebase login won't reuse local data, we enforce a `logout` before each sign in/sign up attempt.


Pascal-Delange (approved):
👍

PR #378, opened 2024-02-21
chore(sentry): update and add integrations

balzdur:
The goal is to improve Sentry monitoring in order to get more information on the "Fetch TypeError" issues.

Not sure it will help but after diging into Sentry for 2 hours, this is the best I found. Let's move empirically on this.


packages/app-builder/src/entry.server.tsx
```diff
@@ -116,6 +116,7 @@ function handleBrowserRequest(
 Sentry.init({
   dsn: getServerEnv('SENTRY_DSN'),
   environment: getServerEnv('SENTRY_ENVIRONMENT'),
+  integrations: [Sentry.httpIntegration(), Sentry.requestDataIntegration()],
```
Pascal-Delange, line 119: while you're at it, maybe we can slowly start decreasing traceSampleRate

Pascal-Delange (approved):

PR #380, opened 2024-02-22
feat(data): add basic data format component

balzdur:
The goal is to add basic data formatter to enhance users data display such as "trigger object" detail in decision viewer.

The current implem :
- add a schema to parse data at runtime
- add a basic data formatter for convenience (in case we want to reuse in multiple places)

NB: it introduce basic formatter for more than just URL. In the above example, look at url, amount, datetime fields

**Before:**
image

**After:**
image


balzdur:
**About the localizaiton:** 

I agree this is a problem, considering we mostly have french based customers, without any possibility to change the locale for formatting. Reading english translated strings is different from reading english formatted numbers/date. 
In order to allow different combination, we must save a full list of formating options on the user profile settings like in the macOS capture from settings bellow.
image
 
I think this is not something we want to prioritize any time soon, so I propose an alternative: bypass the locale handled by the translation service and use `navigator.languages` user setting. 
Exemple: For people using `fr` as main language in their browser settings, formatting will be adapted to `fr` specification.

With that solution, we may cover multiple localisation for formatting and allow an "easy" switch to Marble user settings in the future.

**About the search :**

I must admit this neither came to my mind before; still I think the `ctrl+F` search feature is not something that should drive the formatting logic here. You can still search for formatted string and this will work. IMO, people tends to search using their localised formating rules. 
We can still apply some specific options in order to prevent some frutstration on "unformatted" search. Example: on number formating
```js
new Intl.NumberFormat('en', { useGrouping: true}).format(12345678.264)
// '12,345,678.264'

new Intl.NumberFormat('en', { useGrouping: false}).format(12345678.264)
// '12345678.264'


new Intl.NumberFormat('fr', { useGrouping: true}).format(12345678.264)
// '12 345 678,264'

new Intl.NumberFormat('fr', { useGrouping: false}).format(12345678.264)
// '12345678,264'
```

PR #382, opened 2024-02-26
Pascal/mar 458 data model links must point to a primary key

Pascal-Delange:
WIP for testing https://github.com/checkmarble/marble-backend/pull/491

- [x] implement the main workflows (create unique field, [un]toggle unique field)
- [x] cleanup hardcoded translations, debug logs
- [x] extend data model repository for the edition methods


packages/app-builder/src/routes/ressources+/data+/createField.tsx
```diff
@@ -292,6 +304,39 @@ export function CreateField({ tableId }: { tableId: string }) {
...
+                          <Checkbox
+                            checked={field.value}
+                            disabled={selectedEnum}
+                            onCheckedChange={(checked) => {
+                              field.onChange(checked);
+                            }}
+                          />
+                        </FormControl>
+                        <FormLabel>
+                          <p>{'Is Unique'}</p>
```
balzdur, line 323: If still needed, translate

packages/app-builder/src/routes/_builder+/data.tsx
```diff
@@ -143,6 +164,22 @@ function TableDetails({
             : t('data:required');
         },
       },
+      {
+        id: 'unicityConstraint',
+        accessorKey: 'unicityConstraint',
+        size: 30,
```
balzdur, line 170: Not enough space for small screen.

A rule of thumb: 
- this settings is the minimal pixel length required for a column (so see it as a `min-width` value)
- the relative comparison between each column `size` is used to allow space accorss columns (when there is enough space to at least, respect the `min-width` constraint)


So in the end :
- check for small screen that the header is not clipped
- adapt if needed the size of other columns to distribute space accross columns

As a last resort, you can use `minSize` and/or `maxSize`

packages/app-builder/src/routes/ressources+/data+/editField.tsx
```diff
@@ -78,11 +85,71 @@ export async function action({ request }: ActionFunctionArgs) {
...
+      disabled: true,
+      reason: 'cannot_untoggle_field_linked',
+    };
+  }
+  return {
+    disabled: false,
+    checked: false,
+    reason: '',
+  };
+}
```
balzdur, line 144: IMO, we shouldn't override `checked` property. The value is hold by the data model.

Still, I think disabling (with reason) part of the UI helps prevent unsuitable actions from the user (which in the end will result in errors thrown by the backend) but :
- I prefer to display real value, so the user (and devs) are not confused when there is a problem
- Normally, this kind of "disable stuff we do not want to allow" should comes after handling the possible errors from the backend (but in this special case, I must admit we can skip it, it's long and cumbersome).

packages/app-builder/src/routes/ressources+/data+/editField.tsx
```diff
@@ -171,6 +255,66 @@ export function EditField({
...
+                        <Checkbox
+                          checked={field.value || fieldIsUniqueOverride.checked}
+                          disabled={fieldIsUniqueOverride.disabled}
+                          onCheckedChange={(checked) => {
+                            field.onChange(checked);
+                          }}
+                        />
+                      </FormControl>
+                      <FormLabel>
+                        <p>{'Is Unique'}</p>
```
balzdur, line 274: to translate

packages/app-builder/src/routes/ressources+/data+/editField.tsx
```diff
@@ -78,11 +85,71 @@ export async function action({ request }: ActionFunctionArgs) {
...
+type fieldIsUniqueOverride = {
+  checked: boolean;
+  disabled: boolean;
+  reason:
+    | 'cannot_toggle_enum_enabled'
+    | 'cannot_toggle_index_pending'
+    | 'cannot_untoggle_field_linked'
+    | 'object_id_must_be_unique'
+    | '';
+};
```
balzdur, line 97: To remove.
In that particular type of usecase, recommandation is to let Typescript infer return type (espectially when this is the return type of a local function)

balzdur (approved):

packages/app-builder/src/routes/ressources+/data+/createField.tsx
Pascal-Delange: ah yes my bad, forgot this one

packages/app-builder/src/routes/ressources+/data+/editField.tsx
Pascal-Delange: alright 👍 
This felt right because it made for proper autocomplete where I was using it below but fine

packages/app-builder/src/routes/ressources+/data+/createLink.tsx
```diff
@@ -300,6 +302,11 @@ export function CreateLink({
                     )}
                   />
                 </div>
+                <p>
+                  A link must point to a unique field in the target table. If
+                  you want to point to another field of the table, please mark
+                  it as unique first.
+                </p>
```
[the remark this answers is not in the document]
Pascal-Delange: oopsies, forgot this one too

packages/app-builder/src/routes/_builder+/data.tsx
Pascal-Delange: oh shit yes, I was developing on the big screen and completely forgot about that.

Pascal-Delange: I'm trying 70 which seems to be about enough when the width is 1350px

packages/app-builder/src/routes/ressources+/data+/editField.tsx
Pascal-Delange: but please tell me again, when is the convention to have an explicit output type vs inferred ? When you don't need to verify that your helper function outputs a specific type ?

packages/app-builder/src/routes/_builder+/data.tsx
```diff
@@ -143,6 +164,22 @@ function TableDetails({
...
+          const msg =
+            cell.row.original.unicityConstraint === 'active_unique_constraint'
+              ? '☑'
+              : cell.row.original.unicityConstraint ===
+                  'pending_unique_constraint'
+                ? '🔄'
+                : '☐';
+          return <p className="text-l text-center">{msg}</p>;
+        },
+      },
```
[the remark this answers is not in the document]
Pascal-Delange: 👍 
any tips on how to get the icon aligned to the middle of the column ?
(I tried `justify-center` but to no avail)

packages/app-builder/src/routes/ressources+/data+/editField.tsx
Pascal-Delange: Ok, I understand your point.
The way I ended up there I think is that I wanted to force-display a checked box if `field.unicityConstraint === 'pending_unique_constraint'`. Probably at first I was returning a value for checked only in this case. But then I was using a possible undefined in the `checked={field.value || fieldIsUniqueOverride.checked}`, and the logic ended up depending on the order of values in the `||` (I think I was getting undefined checkbox values (though still displayed as true I think) in the case `undefined || false`, or maybe the other way around, and it was a headache).

Pascal-Delange: Alright perhaps something like 
```
checked={
  fieldIsUniqueOverride.reason ===
  'cannot_toggle_index_pending'
    ? true
    : field.value
}
```
does the job.

Pascal-Delange: or nevermind, I guess we don't need to override the value at all indeed

packages/app-builder/src/routes/_builder+/data.tsx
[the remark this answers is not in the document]
balzdur: Mmh, my first idea would to put the `Icon` inside a `<div className="w-full h-full flex justify-center items-center>`

packages/app-builder/src/routes/ressources+/data+/editField.tsx
balzdur: In general, TS recommandations advise to let TS infer return type of function. In a lot of cases, it does the job better than an explicit declaration (by doing more discrimination + literral types where people tends to define a `string` for example).

Still, there are some exceptions: defining a public interface which in the end will be used outside of the module, in another pkg / file / etc... can be a good idea, since this is the public contract we ensure with the consumer (so in this special case, we want TS to ensure we respect the contract in our implementation).

Basically the question is more or less this one in my mental model :
- I want TS to check explicitly I respect the public contract I defined for this module: explicit types
- I want TS to check the integration of my function: implicit types (TS does a better job at it + VS Code allow autocomplete so no DX lost)

Pascal-Delange: alright thanks!

packages/app-builder/src/routes/_builder+/data.tsx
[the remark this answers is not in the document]
Pascal-Delange: thanks ! I sort of assumed I could center the object in the column even if it's not included in a parent div

PR #386, opened 2024-02-27
feat(auth): improve error handling

balzdur:
Improve error handling :
- handle popup blocked by navigator
- handle firebase network errors
- handle global errors for `useBackendInfo`


packages/app-builder/src/utils/browser.ts
```diff
@@ -0,0 +1,20 @@
+export function getCurrentBrowser(userAgent: string) {
```
Pascal-Delange, line 1: I'm not expert but I was under the impression that browser names in user agents are notoriously unreliable (because of legacy reasons where browers where pretending to be other browsers etc).
is this fine here ?

Pascal-Delange (approved):

packages/app-builder/src/utils/browser.ts
balzdur: This is a totally mess. But for the use we have, I do not want to include a new library to handle this properly. If we do need other discrimination, I may change my mind, in this case, I tested at least on Safari and it works properly .

Safari is the only browser that actually blocked in PROD our auth modal. I suspect chromium based browsers (chrome + brave works properly on my side) to whitelist Firebase auth modals, since they are auth related. But this is an extrapolation of my experiments (no proof). Still, I forced the error on Chrome, Brave and Safari and we get :
- Chrome: browser chrome
- Safari: browser chrome
- Brave: browser chrome

IMo, this is good enough for that specific use case.

Pascal-Delange: deal!

PR #387, opened 2024-02-27
fix(deployment): check permission in loader

balzdur:
Hotfix to allow non admin to access scenario info

PS: if this is ok for you @Pascal-Delange we can cherry-pick this commit on the current released version


Pascal-Delange:
ok to cherry pick

PR #393, opened 2024-03-05
feat(MenuWithCombobox): add MenuWithCombobox component

balzdur:
First PR required for the refactoring of operand edit : introduce a new component Menu + Combobox (in order to drop the quick & dirty previous composition which was hacky).

PS: look at the storybook to see some example (behaviours: especially keyboard + focus managment + dismissing management)

### DIFF
--- a/packages/ui-design-system/src/MenuWithCombobox/MenuWithCombobox.tsx
+++ b/packages/ui-design-system/src/MenuWithCombobox/MenuWithCombobox.tsx
@@ -1,0 +1,195 @@
+import * as Ariakit from '@ariakit/react';
+import clsx from 'clsx';
+import * as React from 'react';
+
+const WithComboboxContext = React.createContext(false);
+const WithComboboxListContext = React.createContext(false);
+const ParentContext = React.createContext(false);
+const HideAllContext = React.createContext<(() => void) | null>(null);
+
+function HideAllContextProvider({ children }: { children: React.ReactNode }) {
+  const hideAll = Ariakit.useMenuContext()?.hideAll ?? null;
+  return (
+    <HideAllContext.Provider value={hideAll}>
+      {children}
+    </HideAllContext.Provider>
+  );
+}
+
+export interface MenuRootProps {
+  searchValue?: string;
+  onSearch?: (value: string) => void;
+  children: React.ReactNode;
+}
+
+/**
+ * A menu component.
+ *
+ * @example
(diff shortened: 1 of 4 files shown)


packages/ui-design-system/src/MenuWithCombobox/MenuWithCombobox.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const parent = React.useContext(ParentContext);
+  const withCombobox = searchValue !== undefined || onSearch !== undefined;
+
+  const element = (
+    <Ariakit.MenuProvider
+      showTimeout={0}
+      placement={parent ? 'right-start' : 'bottom-start'}
+    >
+      <WithComboboxContext.Provider value={withCombobox}>
+        <HideAllContextProvider>{children}</HideAllContextProvider>
```
Pascal-Delange, line 57: what does this do exactly? Hide context providers from screen readers ?

packages/ui-design-system/src/MenuWithCombobox/MenuWithCombobox.tsx
```diff
@@ -0,0 +1,200 @@
+import * as Ariakit from '@ariakit/react';
```
Pascal-Delange, line 1: is this an alternative to radix ?

packages/ui-design-system/src/MenuWithCombobox/MenuWithCombobox.tsx
balzdur: This is a context I created in order to inject a hideAll method.
This allows to close all menu and submenu at once

Pascal-Delange (approved):

packages/ui-design-system/src/MenuWithCombobox/MenuWithCombobox.tsx
balzdur: Yes, TBH I prefer a little bit the simplicity of Radix (composition first with hidden complexity) but Radix lag behind in term of new features and customization is sometimes really hard.

With that new pkg it's simpler (and there is a lot of concrete examples to allow complex components creation). I already used it for the SelectWithCombobox.

As an example, Radix don't have yet any combobox, but the open issue on the topic started 3/4 years ago. This is definitely a pain point for me. Composition of existing components is hard in radix but easier on Ariakit

PR #395, opened 2024-03-07
adapt env variables for docker compose

Pascal-Delange:


packages/app-builder/.env.example
```diff
@@ -10,7 +10,8 @@ NODE_ENV=development
 SESSION_SECRET=SESSION_SECRET
 SESSION_MAX_AGE=43200
 
-MARBLE_API_DOMAIN=http://localhost:8080
+MARBLE_API_DOMAIN_CLIENT=http://localhost:8080
```
Pascal-Delange, line 13: The one case where the two values are not equivalent is if the application (api+front) are running locally in docker containers. In this case, the client (browser) can call the API on localhost but the front server cannot

Dockerfile
```diff
@@ -15,6 +15,9 @@ RUN --mount=type=secret,id=SENTRY_AUTH_TOKEN \
       SENTRY_AUTH_TOKEN=$(cat /run/secrets/SENTRY_AUTH_TOKEN) && \
       export SENTRY_AUTH_TOKEN=$SENTRY_AUTH_TOKEN && \
       pnpm run -r build-with-sourcemaps --release $SENTRY_RELEASE
+# Comment above and uncomment below to build the image locally (not in CI)
+# RUN pnpm --filter=app-builder run build
+ENV NODE_ENV=production
```
balzdur, line 20: Honnestly, I think we could have two dockerfile : one for the CI and one for the local build.

packages/app-builder/src/utils/environment.ts
```diff
@@ -106,7 +107,11 @@ export function getServerEnv<K extends keyof ServerEnvVarName>(
 export function getClientEnvVars() {
   return {
     ENV: getServerEnv('ENV'),
-    AUTH_EMULATOR_HOST: getServerEnv('FIREBASE_AUTH_EMULATOR_HOST'),
+    FIREBASE_AUTH_EMULATOR_HOST: getServerEnv(
+      'FIREBASE_AUTH_EMULATOR_HOST',
+    )?.startsWith('http://')
+      ? getServerEnv('FIREBASE_AUTH_EMULATOR_HOST')
+      : 'http://' + getServerEnv('FIREBASE_AUTH_EMULATOR_HOST'),
```
balzdur, line 114: Why do you need this ?

balzdur (approved):

Dockerfile
Pascal-Delange: hmm, do we ?
I'm assuming us as developers will usually develop in VScode not in a local docker image.
On the other hand our open source users will use the same docker image as us in prod.

Pascal-Delange: I mean, we can add an alternative dockerfile for local docker image building when debugging (as I had to do this week) - but I don't expect it to be used much

packages/app-builder/src/utils/environment.ts
Pascal-Delange: ok good question. The reason was that currently we have `http://localhost:9099` as the value for the env variable in the example env (which is not really a hostname). And it was confusing in the docker compose because the backend expects the hostname without http.
But in fact in the doc they put it without the protocol https://firebase.google.com/docs/emulator-suite/connect_auth#admin_sdks so I guess we can just keep it as `localhost:9099` in our example en after all

balzdur: Yes, I think we can harmonize this and make the change once for all (IIRC, you already raised this issue for other env vars).

PR #403, opened 2024-03-15
refactor(EvaluationErrors): improve error messages UI

balzdur:
Improve `EvaluationErrors` handling: there was a lot of specific rendering for that, especially inside aggregation modal
- expose a better abstraction
  - handle errors as an array to give consistency accross screens to display errors
  - the first refactor I did was too specific
- fix a lot of layering issues (especially between `view` and `edit` mode): this is UI only

### DIFF
--- a/packages/app-builder/src/components/Scenario/ScenarioValidationError.tsx
+++ b/packages/app-builder/src/components/Scenario/ScenarioValidationError.tsx
@@ -1,47 +1,25 @@
-import { type EvaluationError } from '@app-builder/models';
-import {
-  adaptEvaluationErrorViewModels,
-  useGetNodeEvaluationErrorMessage,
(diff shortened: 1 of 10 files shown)


packages/app-builder/src/components/Scenario/ScenarioValidationError.tsx
```diff
@@ -1,47 +1,25 @@
...
-  );
-}
 
 export function EvaluationErrors({
-  evaluationErrors,
+  errors,
   className,
 }: {
-  evaluationErrors: EvaluationError[];
+  errors: string[];
```
balzdur, line 7: `EvaluationErrors` should be error formating agnostic (required to handle specific root error formatter)

Pascal-Delange (approved):

PR #404, opened 2024-03-15
Thomas/add decision scenario iteration

balzdur:
In this PR :
- introduce the `scenario_iteration_id` on decision
- add link on both `DecisionList` and `DecisionDetail` (cf screenshots)
- refactor to use the new `rule_detail` (cf backend PR): consume the real rule score modifier, not the result of the rule execution inside `RuleDetail`

Screenshot 2024-03-15 at 16 28 41
Screenshot 2024-03-15 at 16 28 04

Bonus on case detail:
image

### DIFF
--- a/packages/app-builder/src/components/Decisions/DecisionsList.tsx
+++ b/packages/app-builder/src/components/Decisions/DecisionsList.tsx
@@ -29,7 +29,7 @@ type Column =
   | 'score'
   | 'outcome';
 
-interface DecisionVM {
+interface DecisionViewModel {
   id: string;
   createdAt: string;
   scenario: {
@@ -50,7 +50,7 @@ interface DecisionVM {
 
 type DecisionsListProps = {
   className?: string;
-  decisions: DecisionVM[];
+  decisions: DecisionViewModel[];
   columnVisibility?: Partial<Record<Column, boolean>>;
 } & (WithSelectable | WithoutSelectable);
 
@@ -66,7 +66,7 @@ type WithoutSelectable = {
 
 export function useSelectedDecisionIds() {
   const [rowSelection, setRowSelection] = useState({});
-  const getSelectedDecisionsRef = useRef<() => DecisionVM[]>(() => []);
+  const getSelectedDecisionsRef = useRef<() => DecisionViewModel[]>(() => []);
   const getSelectedDecisions = useCallback(
     () => getSelectedDecisionsRef.current(),
     [],
@@ -83,7 +83,7 @@ export function useSelectedDecisionIds() {
   };
 }
 
-const columnHelper = createColumnHelper<DecisionVM>();
+const columnHelper = createColumnHelper<DecisionViewModel>();
 
 export function DecisionsList({
   className,
--- a/packages/app-builder/src/components/Decisions/DecisionsList.tsx
+++ b/packages/app-builder/src/components/Decisions/DecisionsList.tsx
@@ -93,6 +93,34 @@ export function DecisionsList({
         id: 'scenario_name',
         header: t('decisions:scenario.name'),
         size: 200,
+        cell: ({ getValue, row }) => (
+          <Link
+            to={getRoute('/scenarios/:scenarioId', {
+              scenarioId: fromUUID(row.original.scenario.id),
+            })}
+            onClick={(e) => e.stopPropagation()}
+            className="hover:text-purple-120 focus:text-purple-120 font-semibold capitalize text-purple-100 hover:underline focus:underline"
+          >
+            {getValue()}
+          </Link>
+        ),
+      }),
+      columnHelper.accessor((row) => row.scenario.version, {
+        id: 'scenario_version',
+        header: 'Vi',
+        size: 20,
+        cell: ({ getValue, row }) => (
+          <Link
+            to={getRoute('/scenarios/:scenarioId/i/:iterationId', {
+              scenarioId: fromUUID(row.original.scenario.id),
+              iterationId: fromUUID(row.original.scenario.scenarioIterationId),
+            })}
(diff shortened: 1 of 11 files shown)


Pascal-Delange (approved):

packages/app-builder/src/routes/_builder+/decisions+/$decisionId.tsx
```diff
@@ -56,26 +54,39 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
+        databaseAccessors: accessors.databaseAccessors,
+        payloadAccessors: accessors.payloadAccessors,
+        astOperators,
+        dataModel,
+        customLists: custom_lists,
+      }),
+    );
 
-    return json({
+    return defer({
```
balzdur, line 87: Key part of the new commits :
- fetch additional data related to the current decision 
- use `defer` so it do not block the navigation (= start fetching additional data, render with available data and display a loader until data is streamed to the client)

balzdur (commented):
I reworked this PR to take into account the fact we won't extend decision detail DTO

packages/app-builder/src/components/Decisions/DecisionsList.tsx
```diff
@@ -26,9 +29,28 @@ type Column =
   | 'score'
   | 'outcome';
 
+interface DecisionVM {
```
Pascal-Delange, line 32: what does VM stand for ?

Pascal-Delange: ok I think I got it (view model?) but since it's showing up in just 4 places, let's maybe keep it full text ?

packages/app-builder/src/components/Decisions/RulesDetail.tsx
```diff
@@ -75,35 +78,23 @@ export function RulesDetail({
...
-                        customLists={customLists}
-                        triggerObjectType={triggerObjectType}
-                      />
+                    {ruleExecution.description ? (
+                      <Callout variant="outlined">
+                        {ruleExecution.description}
+                      </Callout>
                     ) : null}
+
+                    <Suspense fallback={t('common:loading')}>
```
Pascal-Delange, line 87: just for my understanding: why does this need to be done async ? There's no network call in the component, right ?

packages/app-builder/src/routes/_builder+/decisions+/$decisionId.tsx
Pascal-Delange: oh, this is surprising syntax. So using "defer" here protects us against uncaught exception and the like even without using await or catch above ?

Pascal-Delange: (I mean ok not that surprising but I didn't know this)

Pascal-Delange (approved):

packages/app-builder/src/components/Decisions/RulesDetail.tsx
balzdur: If you look at the component props declaration, we pass `Promise<....>` for astRuleData (this is due to the `defer` usage).
`Suspense` allow to handle asynchronous variables, displaying the `fallback` until promises resolve

balzdur:
**Exception :**
This do not protect against exception like `try ... catch` block. 
If an exception is raised, this trigger the classic behaviour in this case : display the exported (if any) `ErrorBoundary` component from the page (or bubble up if none)

**Await :**
This king of "replace" the `await` statement.
When a navigation/transition occure in Remix, it `await` for all required loaders to resolve before doing anuthing (we already have some latency due to this in some part of the app in staging/prod due to network + real DB). 
Using a `defer` return statement allow to mix data we require to display the page and data we "defer" :
- required data = data needed to display something valuable at first (preventing to have loaders everywhere)
- deferred data = data not needed at first paint or data that take a huge amount of time to fetch

Here we have data that take some time and that is only used in accordion contents (hidden at first) so it makes sense to `defer` it.

ATM, if the promises reject, the page will transition from "decision content" to "error page". If it occures a lot, we can try/catch the error and return a possibly undefined value (= display an error message only where the data is used).
An alternative would be to move the rule detail inside a dedicated `Outlet` and handle errors the classic way with Remix (so the error component will be only displayed in place of the `Outlet` in the final layout)

packages/app-builder/src/components/Decisions/RulesDetail.tsx
balzdur: **Exception :**
This do not protect against exception like `try ... catch` block. 
If an exception is raised, this trigger the classic behaviour in this case : display the exported (if any) `ErrorBoundary` component from the page (or bubble up if none)

**Await :**
This king of "replace" the `await` statement.
When a navigation/transition occure in Remix, it `await` for all required loaders to resolve before doing anuthing (we already have some latency due to this in some part of the app in staging/prod due to network + real DB). 
Using a `defer` return statement allow to mix data we require to display the page and data we "defer" :
- required data = data needed to display something valuable at first (preventing to have loaders everywhere)
- deferred data = data not needed at first paint or data that take a huge amount of time to fetch

Here we have data that take some time and that is only used in accordion contents (hidden at first) so it makes sense to `defer` it.

ATM, if the promises reject, the page will transition from "decision content" to "error page". If it occures a lot, we can try/catch the error and return a possibly undefined value (= display an error message only where the data is used).
An alternative would be to move the rule detail inside a dedicated `Outlet` and handle errors the classic way with Remix (so the error component will be only displayed in place of the `Outlet` in the final layout)

Pascal-Delange:
Yep that was my understanding 👍 
I meant that it protects against exceptions not caught by the server (because I guess under the hood it's awaited somewhere in "defer")

PR #407, opened 2024-03-21
fix(CaseEvents): accidentally removed comment_added case

balzdur:
Bug spotted during the last demo: the `'comment_added'` case has been accidentally removed


Pascal-Delange (approved):
my bad probably !!

PR #408, opened 2024-03-25
display api key prefix

Pascal-Delange:
with https://github.com/checkmarble/marble-backend/pull/524

Capture d’écran 2024-03-25 à 14 04 35

### DIFF
--- a/packages/app-builder/src/models/api-keys.ts
+++ b/packages/app-builder/src/models/api-keys.ts
@@ -10,16 +10,18 @@ function isApiKeyRole(role: string): role is ApiKeyRole {
 
 export interface ApiKey {
   id: string;
-  organizationId: string;
   description: string;
+  organizationId: string;
+  prefix: string;
   role: ApiKeyRole | 'UNKNWON';
 }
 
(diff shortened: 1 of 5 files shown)


packages/app-builder/src/models/api-keys.ts
```diff
@@ -10,16 +10,18 @@ function isApiKeyRole(role: string): role is ApiKeyRole {
...
   role: ApiKeyRole | 'UNKNWON';
 }
 
 export function adaptApiKey(apiKeyDto: ApiKeyDto): ApiKey {
   const apiKey: ApiKey = {
     id: apiKeyDto.id,
-    organizationId: apiKeyDto.organization_id,
     description: apiKeyDto.description,
+    organizationId: apiKeyDto.organization_id,
+    prefix: apiKeyDto.prefix + '*************',
```
balzdur, line 24: Either store it as `value` or only set the `prefix` (you can still insert `'*******'` when formating the value)

Both solution are fine to me

balzdur (approved):
I think I got my answer for Api key prefix usage 😛

PR #410, opened 2024-03-26
Thomas/display decision rule exec errors

balzdur:
In this PR :
- change the API to handle new `NodeEvaluationDto`
  - change return value model
  - add new runtime error codes
  - extract `NodeEvaluation` models to its own file
- rework CSS so `viewer` can also handle error border variants
- inject evaluation to get errors in `useAstEditor()`

image


packages/app-builder/public/locales/en/scenarios.json
```diff
@@ -156,6 +156,14 @@
...
   "validation.evaluation_error.function_error_one": "function contains errors",
   "validation.evaluation_error.function_error_other": "{{count}} functions contain errors",
+  "validation.evaluation_error.null_field_read_one": "unexpected NULL field read",
+  "validation.evaluation_error.null_field_read_other": "{{count}} unexpected NULL field reads",
+  "validation.evaluation_error.no_rows_read_one": "data point is missing",
+  "validation.evaluation_error.no_rows_read_other": "{{count}} data points are missing",
+  "validation.evaluation_error.division_by_zero_one": "division by zero",
+  "validation.evaluation_error.division_by_zero_other": "{{count}} divisions by zero",
+  "validation.evaluation_error.payload_field_not_found_one": "payload field not found",
+  "validation.evaluation_error.payload_field_not_found_other": "{{count}} payload fields not found",
```
balzdur, line 166: Would be good to review, I tried to give "meaningfull" error messages (still, part of them are still verry techy) but at least it would be great to validate my "personal interpretation"

packages/app-builder/src/components/Decisions/RulesDetail.tsx
```diff
@@ -188,6 +192,7 @@ function RuleFormula({
 }) {
   const astEditor = useAstBuilder({
     backendAst: formula,
+    backendValidation: evaluation,
```
balzdur, line 195: This is the core part: inject evaluation so we get errors handled the same way as validation

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/AggregationEdit/Modal.tsx
```diff
@@ -3,14 +3,16 @@ import {
...
   NewConstantAstNode,
 } from '@app-builder/models';
 import {
   type AggregatorOperator,
   aggregatorOperators,
 } from '@app-builder/models/editable-operators';
+import {
+  computeValidationForNamedChildren,
+  type EvaluationError,
+} from '@app-builder/models/node-evaluation';
```
balzdur, line 15: There is a lot of noise like this (I try to move out of barel files so there are a lot of import changes)

packages/app-builder/src/services/validation/scenario-validation-error-messages.ts
```diff
@@ -9,7 +9,7 @@ import { assertNever } from 'typescript-utils';
 // Edit this type to handle contextual data for each error code
 export type EvaluationErrorViewModel =
   | {
-      error: 'UNEXPECTED_ERROR';
+      error: 'UNEXPECTED_ERROR' | 'RUNTIME_EXPRESSION_ERROR';
```
balzdur, line 12: I consider `'RUNTIME_EXPRESSION_ERROR'` like `'UNEXPECTED_ERROR'`: meaning we will display the raw error message since those are "not qualified" errors that should rarelly hapen.

balzdur (commented):
You can focus on the files with comments, the rest is noise due to :
- import changes
- CSS changes

packages/app-builder/public/locales/en/scenarios.json
```diff
@@ -156,6 +156,14 @@
   "validation.evaluation_error.field_not_found_other": "{{count}} fields not found",
   "validation.evaluation_error.function_error_one": "function contains errors",
   "validation.evaluation_error.function_error_other": "{{count}} functions contain errors",
+  "validation.evaluation_error.null_field_read_one": "unexpected NULL field read",
+  "validation.evaluation_error.null_field_read_other": "{{count}} unexpected NULL field reads",
```
Pascal-Delange, line 160:
```suggestion
  "validation.evaluation_error.null_field_read_other": "{{count}} unexpected NULL fields read",
```

packages/app-builder/public/locales/en/scenarios.json
```diff
@@ -156,6 +156,14 @@
...
+  "validation.evaluation_error.no_rows_read_one": "data point is missing",
+  "validation.evaluation_error.no_rows_read_other": "{{count}} data points are missing",
+  "validation.evaluation_error.division_by_zero_one": "division by zero",
+  "validation.evaluation_error.division_by_zero_other": "{{count}} divisions by zero",
+  "validation.evaluation_error.payload_field_not_found_one": "payload field not found",
```
Pascal-Delange, line 165: I'm not sure when we have this vs just "null field read" 🤔

Pascal-Delange (approved):

packages/app-builder/public/locales/en/scenarios.json
balzdur: I've looked into it, this is basically JSON specific (I guess):
- if key is `undefined` -> ErrPayloadFieldNotFound
- if key is `null`-> ErrNullFieldRead

PR #411, opened 2024-03-27
Thomas/refactor ast data

balzdur:
Refactoring PR :
- remove dead code
- extract ast data (used to define options in Ast Editor and/or display node info) to dedicated contexts
- inject explicit dependencies instead of global `builder` variable

The global goal is to split concerns into separated services (here, we extract contextual data from `useAstBuilder` which is too complicate).

Using contexts help to reduce props drilling in this perticular huge nested components implementation.


Pascal-Delange (approved):
blind approve

PR #415, opened 2024-04-02
Thomas/fuzzy match functions edit

balzdur:
In this PR : first implementation of fuzzy match edit
- handle dynamically conversion to `FuzzyMatch` or `FuzzyMatchAnyOf` (based on selected operands info)
- edit existing nodes
- create a new fuzzy match node (using the `Function` discovery submenu)

New in this PR:
- do not convert to a specific view model like other modals (just narrow the type based on what we know from fuzzy match contract) : easier to reuse existing components
- inject a currated list of available options in operand edit (here: we only allow to select "String" operand for example)
- display list of string/number constants datatype
  - same icon as non list datatype
  - add a specific translation to inform the user it's a key (displayed in the tootlip)

Caveats :
- We do not edit "threshold" (for now, it's initialize to 0).
- Constant coercion is non customizable (= we still can inject non string constant but we can live with that since we do some validation)
- Missing external doc link (especially to give info on algorithms)
- The full list of algo is displayed ATM
- No specific tooltip ATM (we may want to display selected alog and/or link to external doc)

output


packages/app-builder/src/components/Scenario/AstBuilder/AstBuilder.tsx
```diff
@@ -45,14 +46,16 @@ export function AstBuilder({
...
           <AggregationEditModal>
-            <RootAstBuilderNode
-              setOperand={setOperand}
-              setOperator={setOperator}
-              appendChild={appendChild}
-              remove={remove}
-              editorNodeViewModel={editorNodeViewModel}
-              viewOnly={viewOnly}
-            />
+            <FuzzyMatchComparatorEditModal>
```
balzdur, line 49: This is anoying and non scallable.

The current pattern is : 
- use a single modal for each specific edit
- share a context
- use it when required

With that we have issues when :
- nesting modals: we only share a single modal comp for each specific node type (meaning we can't nest them)
- composing modals: depending on which modal need which other modal, the order of composition declared in this file matter.


The long term better plan would be to create a modal locally, for each specific place we need one.

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/AggregationEdit/Modal.tsx
```diff
@@ -207,7 +207,7 @@ export function AggregationEditModal({
     <ModalV2.Root open={open} setOpen={onOpenChange}>
       <AggregationEditModalContext.Provider value={editAgregation}>
         {children}
-        <ModalV2.Content size="large">
+        <ModalV2.Content size="large" unmountOnHide>
```
balzdur, line 210: Required to ensure we refresh at modal opening the initial data (without that, data from one edit to the other can leak: horrible when it's two different nodes.

Using "local modals" would solve this issue by design

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/FuzzyMatchComparatorEdit/Modal.tsx
```diff
@@ -0,0 +1,285 @@
...
+function FuzzyMatchComparatorEditModalContent({
+  initialValue,
+  onSave,
+}: {
+  initialValue: FuzzyMatchEditorNodeViewModel;
+  onSave: (vm: FuzzyMatchEditorNodeViewModel) => void;
+}) {
+  const { t } = useTranslation(['scenarios', 'common']);
+  const getNodeEvaluationErrorMessage = useGetNodeEvaluationErrorMessage();
+  const [value, setValue] = useState<FuzzyMatchEditorNodeViewModel>(
```
Pascal-Delange, line 152: could we `value`=>`thisNode` or something similar ?

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/FuzzyMatchComparatorEdit/Modal.tsx
```diff
@@ -0,0 +1,285 @@
...
+              }
+              options={leftOptions}
+            />
+            <div className="border-grey-10 bg-grey-02 flex h-10 w-fit min-w-[40px] items-center justify-center rounded border p-2 text-center">
+              <span className="text-s text-grey-100 font-medium">~</span>
+            </div>
+            <Operand
+              operandViewModel={value.children[1]}
+              onSave={(ast) => {
+                const nextChild = adaptEditorNodeViewModel({
```
Pascal-Delange, line 206: why is this "nextChild" ? 🤔

packages/app-builder/src/models/data-model.ts
```diff
@@ -187,9 +187,12 @@ export function getDataTypeIcon(dataType?: DataType): IconName | undefined {
     case 'Timestamp':
       return 'schedule';
     case 'String':
+    case 'String[]':
```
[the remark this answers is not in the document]
balzdur: I think I did not get your point here.
Is `getDataTypeIcon` the issue ? or the fact it's present in the file `data-model.ts` ? or something else ?

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/FuzzyMatchComparatorEdit/Modal.tsx
balzdur: The `onSave` we need to pass here is the one responsible for updating the "right" operand (in `left ~ right`).

In order to build this `onSave` specific callback, we have the global `setValue` that require to update the "parent" AST node (not just the `right` operand). I just named it `nextChild` because this is the "future" child after the update, I can change it to `right` or `child`, it's just a habit I add by building some callback where I needed to manipulate "old" or "current" and "future" or "next" value.

packages/app-builder/src/models/data-model.ts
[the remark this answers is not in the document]
Pascal-Delange: Yea it's more the principle that "data model" for me represents the user's data model (as in the schema of his tables), rather than the types of data that are possibly handled in intermediate AST steps. Two things that are very related but one is included in the other.
But that's fine now really

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/FuzzyMatchComparatorEdit/Modal.tsx
Pascal-Delange: All right, gotcha.
Maybe `new` rather than `next` ? I was trying to interpret "next" as "are we iterating over something?".
But it's really a detail, you can ignore this if you want

PR #416, opened 2024-04-02
fix: custom list value deletion was broken

Pascal-Delange:


packages/app-builder/src/routes/_builder+/lists+/$listId.tsx
```diff
@@ -57,7 +57,7 @@ export default function Lists() {
 
   const columns = useMemo(
     () => [
-      columnHelper.accessor((row) => row.value, {
+      columnHelper.accessor((row) => row, {
```
Pascal-Delange, line 60: any chance there is another function to call here since the "identity" first argument is kinda useless ?

balzdur (approved):

PR #419, opened 2024-04-05
Thomas/fuzzy match functions threshold

balzdur:
Final steps to fully implement fuzzy match edit view :
- add threshold edit
- refactor code to extract state logic in a separate file

### DIFF
--- a/packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/FuzzyMatchComparatorEdit/Modal.tsx
+++ b/packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/FuzzyMatchComparatorEdit/Modal.tsx
@@ -1,77 +1,39 @@
 import { Callout } from '@app-builder/components';
 import { ExternalLink } from '@app-builder/components/ExternalLink';
 import { EvaluationErrors } from '@app-builder/components/Scenario/ScenarioValidationError';
-import { type AstNode, isFuzzyMatchComparator } from '@app-builder/models';
-import { type FuzzyMatchAlgorithm } from '@app-builder/models/editable-operators';
-import { type EvaluationError } from '@app-builder/models/node-evaluation';
 import {
-  useAdaptEditableAstNode,
-  useOperandOptions,
-} from '@app-builder/services/ast-node/options';
+  type AstNode,
+  NewFuzzyMatchComparatorAstNode,
+} from '@app-builder/models';
 import {
-  adaptAstNodeFromEditorViewModel,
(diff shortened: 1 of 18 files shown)


packages/app-builder/src/models/fuzzy-match.ts
```diff
@@ -0,0 +1,88 @@
...
+export type FuzzyMatchAlgorithm = (typeof fuzzyMatchAlgorithms)[number];
+
+export function isFuzzyMatchAlgorithm(
+  value: string,
+): value is FuzzyMatchAlgorithm {
+  return (fuzzyMatchAlgorithms as ReadonlyArray<string>).includes(value);
+}
+
+export function getFuzzyMatchAlgorithmName(
+  t: TFunction<['common', 'scenarios'], undefined>,
```
Pascal-Delange, line 21: in the end t is not used, right ?

Pascal-Delange: don't we want to store those in the translation json files ?

Pascal-Delange: (this is where we could use "bag of words similarity" and "full string similarity", or something like this, for more human readable names)

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/FuzzyMatchComparatorEdit/Modal.tsx
```diff
@@ -300,3 +225,75 @@ function FuzzyMatchComparatorEditModalContent({
...
+
+const funcNameTKeys = {
+  FuzzyMatch: 'scenarios:edit_fuzzy_match.fuzzy_match',
+  FuzzyMatchAnyOf: 'scenarios:edit_fuzzy_match.fuzzy_match_any_of',
+} satisfies Record<
+  FuzzyMatchEditorNodeViewModel['funcName'],
+  ParseKeys<['scenarios']>
+>;
+
+function Examples({
```
Pascal-Delange, line 237: This is really pretty cool 👌  great idea on this

Pascal-Delange (approved):
cool!

packages/app-builder/src/models/fuzzy-match.ts
balzdur: I added translation for both editable algorithm.
Just check I did not associated wrong translation to each algorithm, and if it's OK, this is mergeable.

Pascal-Delange: looks good !

PR #422, opened 2024-04-09
feat(marble-api): handle breaking changes in marble-api

balzdur:
Adapt the frontend to the breaking change on DTOs

### DIFF
--- a/packages/marble-api/scripts/openapi.yaml
+++ b/packages/marble-api/scripts/openapi.yaml
@@ -125,7 +125,7 @@ paths:
             items:
               type: string
         - name: has_case
-          description: 'Optional filter: the decision is attached to a case'
+          description: Filter decisions that have a case associated with them or not (true or false, default returns all)
           in: query
           required: false
           schema:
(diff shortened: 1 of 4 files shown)


packages/marble-api/scripts/openapi.yaml
```diff
@@ -116,40 +116,44 @@ paths:
...
           in: query
           required: false
           schema:
             type: array
             items:
               type: string
-        - name: startDate
-          description: start date used to filter the list
+        - name: has_case
+          description: 'Optional filter: the decision is attached to a case'
```
Pascal-Delange, line 128:
```suggestion
          description: 'Optional filter: the decision is attached to a case (true/false)'
```

Pascal-Delange (approved):

packages/marble-api/scripts/openapi.yaml
balzdur: Good catch, I propose an alternative description (no filter behaviour is explicit) and copy paste the same description on the `public-api.yaml`

Pascal-Delange (approved):

PR #427, opened 2024-04-17
Thomas/mar 528 dataviz of an organization data model

balzdur:
First iteration on the datamodel dataviz :
- new routing to allow both presentation (classic list + new schema view)
- in the schema:
  - same edit actions as in the classic view (it wasn't difficult to add since we do have existing actions packages as `/ressources/*`)
  - add some "custom action":
    - one for "auto layout"
    - one for "focus selected table"

The main difficult I faced is the layout: since we do use "dynamically sized" components. I ended up with an "opacity hack", meaning every new edge/node is painted on screen with `opacity: 0` (hidden for the user) so it's width + height is computed and a second operation auto-layout everytime a new element is added.


packages/app-builder/src/components/Data/DataModelFlow.tsx
```diff
@@ -0,0 +1,269 @@
...
+      edges.some((ed) => ed.style?.opacity === 0)
+    ) {
+      const layouted = getLayoutedElements(nodes, edges);
+      setNodes(layouted.nodes);
+      setEdges(layouted.edges);
+      window.requestAnimationFrame(() => {
+        fitView();
+      });
+    }
+  }, [fitView, edges, nodes]);
```
balzdur, line 142: This is where the reconciliation logic + opacity hack is implemented.
The rest is just "basic code"

packages/app-builder/src/components/Data/DataModelFlow.tsx
```diff
@@ -0,0 +1,269 @@
...
+  const [edges, setEdges] = React.useState<Array<Edge<DataModelEdgeData>>>([]);
+
+  const onNodesChange = React.useCallback((changes: NodeChange[]) => {
+    const allowedChanges = changes.filter((change) => change.type !== 'remove');
+    setNodes((nds) => applyNodeChanges(allowedChanges, nds));
+  }, []);
+  const onEdgesChange = React.useCallback((changes: EdgeChange[]) => {
+    const allowedChanges = changes.filter((change) => change.type !== 'remove');
+    setEdges((eds) => applyEdgeChanges(allowedChanges, eds));
+  }, []);
```
balzdur, line 86: I prevent here the "remove" action (before that, you could remove nodes and edges despite this is not triggering any real action on the org data model)

packages/app-builder/src/components/Data/DataModelFlow.tsx
```diff
@@ -0,0 +1,269 @@
...
+        position,
+        style: { ...nd.style, opacity: 1 },
+      } satisfies Node<DataModelNodeData>;
+    }),
+    edges: edges.map((ed) => {
+      return {
+        ...ed,
+        style: { ...ed.style, opacity: 1 },
+      } satisfies Edge<DataModelEdgeData>;
+    }),
```
balzdur, line 267: Force the opacity to `1` for all the nodes after a layout

packages/ui-design-system/src/Menu/Menu.stories.tsx
```diff
@@ -13,7 +13,7 @@ import {
   MenuItem,
   MenuPopover,
   MenuRoot,
-} from './MenuWithCombobox';
+} from './Menu';
```
balzdur, line 16: Rename since this was confusing: you can actually use this without a combobox, I started by implementing a new component before figuring out it was already done...

packages/app-builder/src/services/data/data-model.tsx
```diff
@@ -0,0 +1,8 @@
+import { type TableModel } from '@app-builder/models';
+import { createSimpleContext } from '@app-builder/utils/create-context';
+
+const DataModelContext = createSimpleContext<TableModel[]>('DataModelContext');
+
+export const DataModelContextProvider = DataModelContext.Provider;
+
+export const useDataModel = DataModelContext.useValue;
```
balzdur, line 8: New context to mutualize datamodel fetching for both new pages (it's the same pattern we already have for `currentScnearioIteration`)

balzdur (commented):
Focus on the comment for interesting parts

Pascal-Delange (approved):
👍

PR #430, opened 2024-04-23
fix(TimeAddEdit): dropdown should extend outside the modal

balzdur:
I tried to go further in the fix by using the `Operand` shared component as I did inside the "fuzzy match" edit modal but I faced the same issue related to shared modal context...

So for now, those changes fix the bug, and I created Linear tasks to tackle the long term refactor accordingly in the future.


Pascal-Delange (approved):
👍

PR #431, opened 2024-04-23
feat(rules): add global local filter to rules list

balzdur:
Implement default global filter for rules list :
- default React Table implementation (= exact match)
- global (= on all columns)
- local (= no async call to server)

Bonus: reuse "highlight" matching text in purple comp (to improve UX)

output


Pascal-Delange (approved):
nice!

PR #435, opened 2024-04-26
Thomas/mar 534 create UI for a data model pivot

balzdur:
Simple implementation to create a pivot on a data model pivot :
- add menu option inside existing menu (cf screen shot)
- basic modal with select field

Out of scope of this PR:
- basic error handling (ex: narrow error when a pivot already exists)
- conditional rendering based on the presence of an existing pivot

image
image


packages/app-builder/src/services/data/pivot.tsx
```diff
@@ -0,0 +1,91 @@
...
+        baseTableId: tableModel.id,
+        fieldId: field.id,
+        id: field.id,
+        displayValue: field.name,
+      });
+    });
+
+  const tablesMap = new Map(dataModel.map((table) => [table.id, table]));
+
+  function recursiveLinkedPivot(
```
Pascal-Delange, line 47: just a syntax preference here, but I'd suggest making this separately defined recursive function that returns a set/array of possible pivots, rather than a closure defined inside getPivotOptions and mutating the variable `pivots` (I think that makes it easier to read, and we're not really in a case where there is a strong argument for a closure)

Pascal-Delange (commented):
I'm not sure at first glance, but it doesn't seem to include 2nd level options 🤔  (I made a test locally and it only proposed me the first level of links as pivots)

packages/app-builder/src/services/data/pivot.tsx
balzdur: Done.
I also added some basic tests (but it's not easy to read due to the datamodel + links creation)

balzdur: Concerning "2nd level options", it worked on my side (before and after the refactoring).

Can you check your datamodel "2nd level option" is a String field ?
NB: the pivot naming should be a concatenation of links names

Pascal-Delange: seems good now anyway 👍

PR #440, opened 2024-04-30
Thomas/mar 547 display information on decisionsdecisionid about the pivot

balzdur:
Add information on decision detail to ease finding related decisions.

In addition :
- add `field` and `link` tags (to distinguish both pivot types)
- add conditionnal logic to render helping message

NB: when available, I'll update the appropriate doc link. For now, it points to the API endpoint.
output


Pascal-Delange (approved):
👍  let's just add a proper link for pivot definition (I'll send one RN)

PR #441, opened 2024-05-02
Thomas/mar 552 access transfer par transferid

balzdur:
In this PR :
- create the transfercheck repository
- add `partner_id` info on the user
- implement the "transfercheck" app skeleton
- handle "app routing" based on user role

This is not yet ready for a "V0" on transfercheck app but this PR is already too big


packages/app-builder/src/routes/$.tsx
```diff
@@ -0,0 +1,11 @@
+import { serverServices } from '@app-builder/services/init.server';
+import { getRoute } from '@app-builder/utils/routes';
+import { type LoaderFunctionArgs } from '@remix-run/node';
+
+export async function loader({ request }: LoaderFunctionArgs) {
+  const { authService } = serverServices;
+  await authService.isAuthenticated(request, {
+    successRedirect: getRoute('/app-rooter'),
```
Pascal-Delange, line 8:
```suggestion
    successRedirect: getRoute('/app-router'),
```
(ditto file name)

packages/app-builder/src/repositories/TransferRepository.ts
```diff
@@ -0,0 +1,33 @@
+import { type MarbleApi } from '@app-builder/infra/marble-api';
+import { adaptTransfer, type Transfer } from '@app-builder/models/transfer';
+
+export interface TransferRepository {
+  listTransfers(args: {
+    /**
+     * ID of the transfer in the Partner database (do not confuse with the ID generated by Marble database)
```
Pascal-Delange, line 7: TBH if the naming is too ambiguous, we could rename it - for starters in the front end repo openapi and models.

Pascal-Delange (approved):
👍

packages/app-builder/src/repositories/TransferRepository.ts
balzdur: Good idea. I'll change it in the models, but not in the `openapi` for two reasons :
1. I can't change query param naming since this will break the contract
2. I really like to be as closed as possible as backend naming (this convention proved to be really helpfull to iterate over the `openapi.yaml` along the time)

PR #442, opened 2024-05-03
Thomas/mar 552 access transfer par transferid

balzdur:
In this PR :
- fetch transfer based on the procided partner transfer id
- fetch data on the transfer detail page

It also tackle all the edge cases (eg "trasnfer not found", "no transfers corresponding to your search"...)

I'll do the patch on monday

output


Pascal-Delange (approved):
👍

PR #447, opened 2024-05-14
feat(pivot): filter out field pivot options used in link options

balzdur:
Opinionated way to remove possible "duplicate" pivot options (cf Linear issue for an example).

As a reminder, a pivot field is identified by the field name while an "equivalent" link field by the link name. If they both have the same value, this may introduce confusion (like in the example from Linear issue).


Pascal-Delange (approved):
👍

PR #449, opened 2024-05-20
feat(auth): add redirect to for logout

balzdur:
Implement `redirectTo` feature for sign in :
- if `redirectTo` is a safe path (= relative path, only internal to the app)
- then redirect to the provided location

Caveat: I only use this feature for "client" logout comming from the firebase auto-refresh hook. IMO, this is the main source of the problem. If the problem persist, I can look into doing the same for the `isAuthenticated(...)` guard used in loaders/actions.

I prefer to iterate slowly on the auth service since the Firebase integ is kind of hacky with SSR (+ Remix still don't provide serious middleware integration to plug into the server)

### DIFF
--- a/packages/app-builder/src/routes/ressources+/auth+/refresh.tsx
+++ b/packages/app-builder/src/routes/ressources+/auth+/refresh.tsx
@@ -41,7 +41,16 @@ export function useRefreshToken() {
           );
         },
         () => {
-          navigate(getRoute('/ressources/auth/logout'));
+          let redirectUrl = getRoute('/ressources/auth/logout');
+          if (window) {
+            const searchParams = new URLSearchParams();
+            searchParams.set(
+              'redirectTo',
+              `${window.location.pathname}${window.location.search}`,
+            );
+            redirectUrl = `${redirectUrl}?${searchParams.toString()}`;
+          }
+          navigate(redirectUrl);
         },
       );
     },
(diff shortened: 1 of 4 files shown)


packages/app-builder/src/routes/ressources+/auth+/refresh.tsx
```diff
@@ -41,7 +41,16 @@ export function useRefreshToken() {
...
         () => {
-          navigate(getRoute('/ressources/auth/logout'));
+          let redirectUrl = getRoute('/ressources/auth/logout');
+          if (window) {
+            const searchParams = new URLSearchParams();
+            searchParams.set(
+              'redirectTo',
+              `${window.location.pathname}${window.location.search}`,
+            );
+            redirectUrl = `${redirectUrl}?${searchParams.toString()}`;
```
Pascal-Delange, line 51: to be sure I totally understand, what happens here? If you're currently editing something and you get refreshed, you'd still loose your local state, wouldn't you ?

balzdur: Exactly. In fact it depends on the screen : for example, we use query params to store "filters state" in some views. Those screens will maintain "internal" state.

We could ser/de state to either :
- local storage
- client browser navigation state (not visible in the URL)
- specific query param + encoding

But I think this bring a lot of complexity due to load/unload that we do not want to maintain (at least for now)

Pascal-Delange (approved):

PR #452, opened 2024-05-22
Thomas/validate workflows

balzdur:
Implement validation :
- global workflow validation rules
- business logic workflow validation rules

Caveat: for now, validation is giving "global" error messages when a node is not valid (= configuration issue). we could improve error messages by using field per field error messages (like in a form) but it is out of scope of this first iteration IMO

output


packages/app-builder/src/components/Scenario/Workflow/DetailPannel/Checklist.tsx
```diff
@@ -0,0 +1,256 @@
+import type * as React from 'react';
```
Pascal-Delange, line 1: rename folder to DetailPanel (one "n")

Pascal-Delange (approved):

PR #460, opened 2024-05-31
Thomas/link to external doc

balzdur:
First work on linking documentation inside the product.


packages/app-builder/src/components/HelpCenter.tsx
```diff
@@ -0,0 +1,394 @@
...
+          tags: ['IBAN', 'Data'],
+          href: 'https://docs.checkmarble.com/reference/iban-hashing',
+        },
+      ],
+    }),
+    [t],
+  );
+
+  return { defaultTab, resources };
+}
```
balzdur, line 394: This is the content part : ATM, everything is hardcoded and not localised. I think it's unecessary ATM since our doc is in English only.

What could be done : move `href` def to the newly created file (at least links we also want to reference inside the app)

packages/ui-design-system/src/Kbd/Kbd.tsx
```diff
@@ -0,0 +1,13 @@
+import clsx from 'clsx';
+
+export function Kbd({ className, ...props }: React.ComponentProps<'kbd'>) {
```
Pascal-Delange, line 3: what does Kbd stand for?

packages/app-builder/src/components/HelpCenter.tsx
```diff
@@ -0,0 +1,394 @@
...
+        },
+        {
+          label: 'Scenario Decision',
+          tags: ['Outcome', 'Approve', 'Review', 'Decline'],
+          href: 'https://docs.checkmarble.com/docs/decision-1',
+        },
+        {
+          label: 'Formula',
+          tags: ['Rule formula', 'Trigger formula'],
+          href: 'https://dash.readme.com/project/checkmarble/v1.0.0/docs/formula',
```
Pascal-Delange, line 270: careful here, that's the link to our dashboard not the public page

Pascal-Delange (approved):

packages/ui-design-system/src/Kbd/Kbd.tsx
balzdur: This is based on the [HTML tag with the same name](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd). It's a semantic comp that should define keyboard shortcut a user can use to trigger some actions.

PR #462, opened 2024-06-03
Thomas/inline links to external doc

balzdur:
In this PR :
- update some links inside the `HelpCenter`
- add inline links to some key part of the app (to help user find appropriate documentation on complex part of the app)

**some screenshots to help visualize changes**

On Scenario, add doc link + change `Callout` variant (based on a previous discussion, personally this lack a little bit of contrast to me but I'm colorblind so I need some second advice on this)
image
image

In operand edit modals:
image
image
image

In pivot creation modal:
image

Other places: workflow, upload csv data


Pascal-Delange (approved):
the contrast works well 👍 
I'm rather a bit bugged by the fact that the arrow feels "too high" compared to the text, but honestly that's probably fine

balzdur:
👍🏽 Je vais remove la flèche, on voit suffisament comme ça que c'est un link

PR #474, opened 2024-06-17
fix(segment): use zod schema with safeParse

balzdur:
Refactor analytics :
- use zod schema to ensure consistant data collection (and re-use already defined custom schema)
- use `safeParse` to only collect valid data (and silently ignore parsing failure)

NB: this is motivated by [this](https://checkmarble.sentry.io/issues/4573421844/?environment=production&project=4506060678037504&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=90d&stream_index=0&utc=true) recent Sentry unhandled error


Pascal-Delange (approved):
good idea 👍

PR #477, opened 2024-06-20
Thomas/protect premium features

balzdur:
In this PR, handle the user roles premium feature :
- harmonize roles api and models declaration (for user & inbox user)
- handle dynamic roles list based on the license

### DIFF
--- a/packages/app-builder/src/services/feature-access.server.ts
+++ b/packages/app-builder/src/services/feature-access.server.ts
@@ -142,5 +142,19 @@ export function makeFeatureAccessService({
     }) => {
       return userPermissions.canEditInboxes;
     },
+    getUserRoles: async () => {
+      const licenseEntitlements = await getLicenseEntitlements();
+      if (licenseEntitlements.userRoles) {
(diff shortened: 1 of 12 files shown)


packages/app-builder/src/services/feature-access.server.ts
```diff
@@ -142,5 +142,19 @@ export function makeFeatureAccessService({
...
+      }
+      return ['ADMIN'] as const;
+    },
+    getInboxUserRoles: async () => {
+      const licenseEntitlements = await getLicenseEntitlements();
+      if (licenseEntitlements.userRoles) {
+        return ['admin', 'member'] as const;
+      }
+      return ['admin'] as const;
+    },
```
balzdur, line 158: I wasn't sure about this.

The alternative would have to expose a `isUserRolesAvailable` method like other premium features, and move roles getter in another place.

Pascal-Delange: I see,
I don't have a strong opinion on this one, so since you already wrote it like this let's keep it 👍

Pascal-Delange (approved):

PR #483, opened 2024-06-26
Thomas/transfercheck api

balzdur:
Split out transfercheck API and add partner repository to fill a `TODO` in the user modal view.

image

Next steps: to better split out dependencies, we still need :
- to create a dedicated auth api
- to create dedicated repositories + services initializers


Pascal-Delange (approved):
👍

PR #484, opened 2024-06-26
feat(api): add alert endpoints for transfer check

balzdur:
Implement alert endpoints for transfercheck

### DIFF
--- a/packages/app-builder/src/repositories/TransferAlertRepository.ts
+++ b/packages/app-builder/src/repositories/TransferAlertRepository.ts
@@ -1,0 +1,66 @@
+import { type TransfercheckApi } from '@app-builder/infra/transfercheck-api';
+import {
+  adaptCreateTransferAlertDto,
+  adaptTransferAlert,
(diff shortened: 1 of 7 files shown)


packages/app-builder/src/repositories/TransferAlertRepository.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      const { alert } = await transfercheckApi.updateAlert(
+        alertId,
+        adaptUpdateTransferAlertDto(updateTransferAlert),
+      );
+
+      return adaptTransferAlert(alert, partnerId);
+    },
+  });
+}
```
balzdur, line 66: the runtime requirement of `partnerId` is a "hack" necessary since we services initialization is "app agnostic" (= we init a `TransferAlertRepository` even for non transfercheck users, so no `partnerId` for them... Doing the `invariant` check is a way to ensure in runtime we do have the required info)

Pascal-Delange (approved):
👍

PR #487, opened 2024-07-02
refactor(alerts): split sent and received alerts and add filters

balzdur:
Huge PR to introduce :
- split between sent and received alerts
- implement filters on the alerts list
  - full text search on message
  - status filter
  - created_at filter (= date range)

NB: alert detail split will come in a future PR and proper alert status also (currently it's the old `read`, `unread`, `archived`)

image


packages/app-builder/src/routes/transfercheck+/alerts+/inboxes._index.tsx
```diff
@@ -0,0 +1,11 @@
+import { serverServices } from '@app-builder/services/init.server';
```
Pascal-Delange, line 1: I'm perhaps nitpicking here but "inboxes" seems a bit weird considering half of it is rather an "outbox".
How about `alerts+/list` instead?

Pascal-Delange (approved):

PR #496, opened 2024-07-05
refactor(transfer-alert): improve validation

balzdur:
Small changes :
- better schema (only move optionnal + max length, based on what we discussed I didn't implemented a "smart" schema on IBAN or Transfer end2end id requirement)
- consistent ordering based on the same order used in the AlertData component
- small fixes on the adapters to allow the possibility to remove an optionnal field (with the current optionnal implem on the API, a missing field is equivalent to no PATCH changes, so we need to explicitly inform of removed fields)

NB: in fine, in a lot of places, I tends to inject "nullable values" for all missing fields to enforce the possibility to remove a field. Finally, I do more or less a PUT every time... not a big deal but if you have a better idea on how do deal with PATCH +  "delete some fileds" tell me your idea

### DIFF
--- a/packages/app-builder/src/models/transfer-alert.ts
+++ b/packages/app-builder/src/models/transfer-alert.ts
@@ -5,6 +5,7 @@ import {
   type TransferAlertUpdateAsBeneficiaryBodyDto,
   type TransferAlertUpdateAsSenderBodyDto,
 } from 'marble-api/generated/transfercheck-api';
+import { z } from 'zod';
 
 export const transferAlerStatuses = [
   'pending',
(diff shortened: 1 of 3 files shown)


Pascal-Delange:
> NB: in fine, in a lot of places, I tends to inject "nullable values" for all missing fields to enforce the possibility to remove a field. Finally, I do more or less a PUT every time... not a big deal but if you have a better idea on how do deal with PATCH + "delete some fileds" tell me your idea

Yep I'm afraid I don't have a silver bullet either... IMO the best workable solution is a nullable input where null input = ignore it, empty string input = reset it

Pascal-Delange (approved):

PR #498, opened 2024-07-17
Thomas/webhooks

balzdur:
Verry basic skeleton to manage webhooks :
- list
- create
- update
- delete

Only URL is edittable for now.

NB: if the URL is "unreachable" the backend (due to Convoy) return an error. If we plan to give usefull feedback on this to the user, I need to get structured error return from the backend (so I can display conditionnal message to the user)

output


packages/app-builder/src/models/webhook.ts
```diff
@@ -0,0 +1,116 @@
...
+  type WebhookDto,
+  type WebhookRegisterBodyDto,
+  type WebhookSecretDto,
+  type WebhookUpdateBodyDto,
+  type WebhookWithSecretDto,
+} from 'marble-api/generated/marblecore-api';
+import invariant from 'tiny-invariant';
+
+export const eventTypes = ['case_status_updated'] as const;
+type EventType = (typeof eventTypes)[number];
```
Pascal-Delange, line 11: can you please explain this syntax to me ? 🤔

Pascal-Delange (approved):
all clear to me ! 
One question: do we really want to allow modifying the url to a webhook ? (as opposed to require to create a new endpoint)

packages/app-builder/src/models/webhook.ts
balzdur: This is TS syntax: basically it create a type from all the elements of the "enumerable" provided type (here, union of each `eventTypes` items type).

Since we use a `as const` declaration, TS don't infer  :
```TS
const eventTypes: string[]
type EventType = string
```
But
```TS
const eventTypes: readonly ["case_status_updated"]
type EventType = "case_status_updated"
```

Pascal-Delange: thanks,
yes that was my intuition, but why the `[number]` ?

balzdur: I don't know 🤷🏽 
Probably the syntax decision in TS to tell you want to "get the value of the array type by iterating with any number" (since indexes are numbers)

PR #499, opened 2024-07-19
Thomas/update conform

balzdur:
Finally did it, it was painfull but here we are !

In short: I updated to the last major of `conformr` (= the form lib I want to use in the app, that should replace entirely `react-hook-form` in the end).

NB: I tested all the forms to make sure no regressions are introduced + improved existing one (ex: when creating a scenario, error messages are displayed now). 
Another UX improvment, when an error occure at first submit, each fields is validated "on change", so no more "dangling" error messages above fields (when the validation can be done frontend side, of course).

I honnestly doubt you can review it... reading some forms update should be a good way to catch the main changes :
- `FormProvider` to access the state in order to give progressive enhancement
- `Globally the same interface we already used (for `FormField`, `FormInput`....)
- A better API to manipulate form data server side (possibility to return server errors directly on the form, huge improvement of the DX on this)
- Totally new way to use custom components in forms


package.json
```diff
@@ -13,13 +13,13 @@
   },
   "devDependencies": {
     "@vitejs/plugin-react": "4.3.1",
-    "@vitest/coverage-v8": "^1.6.0",
-    "@vitest/ui": "1.6.0",
-    "prettier": "^3.3.2",
+    "@vitest/coverage-v8": "^2.0.3",
```
Pascal-Delange, line 16: just to be sure and since those are major version changes, can you please double check the license type didn't change ?

packages/app-builder/package.json
```diff
@@ -35,13 +35,13 @@
...
     "tsx": "^4.15.7",
-    "vite": "^5.3.1"
+    "vite": "^5.3.4"
   },
   "dependencies": {
     "@ariakit/react": "^0.4.7",
-    "@conform-to/dom": "^0.9.1",
-    "@conform-to/react": "^0.9.1",
-    "@conform-to/zod": "^0.9.1",
+    "@conform-to/dom": "^1.1.5",
```
Pascal-Delange, line 42: ditto here

packages/app-builder/src/components/Form/FormCheckbox.tsx
```diff
@@ -1,24 +1,32 @@
...
   children,
   onCheckedChange,
   ...rest
-}: Omit<React.ComponentProps<typeof Checkbox>, 'checked'>) {
-  const config = useFieldConfig();
+}: FormCheckboxProps) {
+  const name = useFieldName();
+  const [meta] = useField<boolean>(name);
+
+  const control = unstable_useControl(meta);
```
Pascal-Delange, line 17: sanity check: what should I make of "unstable_" ?

Pascal-Delange (approved):
For my culture I think I'm still interested in an oral explanation of the changes to the components/Form files - otherwise LGTM

packages/app-builder/package.json
balzdur: I think the comment is not at the good position

packages/app-builder/src/components/Form/FormCheckbox.tsx
balzdur: Nothing 🤷🏽 

This is a common convention for pkg mainteners to tell the public API is unstable and may be changed with breaking changes in the future; still the hook is introduced for more than 4 months now, w\ any changes. This was created in order to fix issues with the initial attempt of handling form custom inputs `useInputControl`.

More info [here](https://github.com/edmundhung/conform/releases/tag/v1.0.2)

PR #502, opened 2024-07-23
Thomas/webhooks UI

balzdur:
Handle all fields in webhook mutation :
- add event types to webhooks list
- handle all fields in create/update modal

NB: currently the URL is still editable from the UI, I needed to do a workarround in an adapter due to Convoy returning default 0 for optionnal timeout & co...

output


packages/app-builder/src/components/Form/FormErrorOrDescription.tsx
```diff
@@ -0,0 +1,59 @@
...
+          descriptionClassName,
+        )}
+        {...props}
+      >
+        {description}
+      </p>
+    );
+  }
+  return null;
+});
```
balzdur, line 59: Refactor to adapt the following convention :
- if any error, display the error
- if no error, and a description is provided : display the description
- else display nothing

This allow to add fields description above "interactable" (eg input) and not rely on the placeholder.
This is usefull for certain type of input (ex: number input, which rarelly display a placeholder as soon as you give a default value (like 0))

packages/app-builder/src/components/Form/FormInput.tsx
```diff
@@ -28,15 +28,18 @@ export const FormInput = React.forwardRef<
...
   return (
     <Input
       ref={ref}
       borderColor={meta.valid ? 'grey-10' : 'red-100'}
       {...inputProps}
-      {...getInputProps(meta, { type })}
+      {...getInputProps(meta, {
+        type,
+        ariaDescribedBy: description ? meta.descriptionId : undefined,
+      })}
```
balzdur, line 42: I need to handle `desccription` at the field level so `getInputProps(...)` can adapt the `aria-described-by` html attribute dynamically

balzdur (commented):
A lot of changes due to a refacto. You can focus on :
- what i commented (for the global changes)
- webhooks related files
  - webhooks+/create
  - webhooks+/update
  - settings+/webhooks

packages/app-builder/src/components/Form/FormErrorOrDescription.tsx
Pascal-Delange: got it

Pascal-Delange (approved):

PR #515, opened 2024-08-01
Thomas/ligther layout

balzdur:
Too much non commited work on local, difficult to keep mooving 😅 

So here is an part of the work I have locally: mostly reduce the layout to prepare for inline decision detail view (TODO part in the above video)

output


pascaldelange (approved):
👍 
purely on the design side, I have a hunch that we may want to make the comments history more quickly accessible - that can be for another PR of course

PR #521, opened 2024-08-08
new case event: rule snooze created

pascaldelange:
Goes with https://github.com/checkmarble/marble-backend/pull/643.
Work is not finished: I'm not happy with the level of detail we're able to display. To discuss together.


packages/app-builder/src/components/Cases/CaseEvents.tsx
```diff
@@ -107,6 +107,7 @@ export function getEventIcon(event: CaseEvent) {
         />
       );
     case 'decision_added':
+    case 'rule_snooze_created':
```
pascaldelange, line 110: going to put another icon here

balzdur:
Taken into account in another PR

PR #522, opened 2024-08-08
Thomas/rule snoozing case event

balzdur:
In this PR :
- add new RuleSnooze repository
- add optionnal comment on rule snooze creation
- handle new case event type (for rule snooze creation)
- add new webhook event type

output

> NB: the gif is taken with a "slow 3G" network throttle to visualize the loading state


pascaldelange (approved):
cool ! thanks

PR #523, opened 2024-08-09
Thomas/demo feedbacks

balzdur:
Feedbacks from the last demo

### DIFF
--- a/packages/app-builder/src/components/Cases/CaseEvents.tsx
+++ b/packages/app-builder/src/components/Cases/CaseEvents.tsx
@@ -431,12 +431,12 @@ function RuleSnoozeDetail({ ruleSnoozeId }: { ruleSnoozeId: string }) {
         className="text-grey-100 text-s"
         {...(isLoading
           ? {}
-          : getCopyToClipboardProps(data.ruleSnooze.pivotValue))}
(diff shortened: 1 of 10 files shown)


packages/app-builder/src/components/Cases/CaseEvents.tsx
```diff
@@ -144,6 +151,16 @@ export function getEventIcon(event: CaseEvent) {
   }
 }
 
+export function getDefaultOpen(event: CaseEvent) {
+  const { eventType } = event;
+  switch (eventType) {
+    case 'comment_added':
```
pascaldelange, line 157: Plus rule_snooze_added, since it contains a message ? (or whatever id we gave it)

pascaldelange (approved):

packages/app-builder/src/components/Cases/CaseEvents.tsx
balzdur: I'll do it, we just need to keep in mind doing so will result in a 3 API call per rule snooze to the backend. Unfortunately, I need to get :
- the snooze detail
- the rule detail (to get scenario iteration id + rule name)
- the scenario iteration detail (to get the scenario id)

This is non bloquant for the frontend navigation to occure (those requests are done lazilly) but this may spam the backend. Still, based on our current usage, this should not be a problem for now.

PR #524, opened 2024-08-09
refactor: remove intermediate context to prevent flaky errors

balzdur:
In this PR, I implement the rule snoozing feedback on "activation" :
- add contextual data
- bonnus: delete intermediate context

We have flaky errors from time to time (Sentry + local) when changing version arround like :
- changing using the version picker
- creating a new draft

Example of error : `No rule corresponding to ${ruleId}`

I suspect this is due to React 18 concurrent mode :
- Data is fetched by Remix by loaders on navigation change (new scenario version to fetch)
- Data should be injected in shared context I created (to ensure consumers are wrapped inside a Context provider, something `useRouteLoaderData` do not ensure)
- Some data are derived from the context, using parameters (like the one picking the rule and raising the above error: use th scenario iteration data + ruleId from URL params)

If the context data is not updated "fast" enough compared to the URL: old rules + new ruleId => error

Using the primitive hook from Remix for "loaders" data may solve the problem if they use internally `useTransition` hook from React (to ensure concurrent mode consider navigation updates as atomic ones = data fetching + rerender).

output


packages/ui-design-system/src/Modal/Modal.tsx
```diff
@@ -5,7 +5,7 @@ import clsx from 'clsx';
 import { forwardRef } from 'react';
 
 const modalContentClassnames = cva(
-  'bg-grey-00 fixed left-1/2 top-1/2 flex h-fit w-full -translate-x-1/2 -translate-y-1/2 flex-col rounded-lg drop-shadow-xl overflow-hidden',
+  'bg-grey-00 fixed left-1/2 top-[10vh] flex h-fit w-full -translate-x-1/2 flex-col rounded-lg drop-shadow-xl overflow-hidden',
```
balzdur, line 8: Required to fix a weird bug, with CSS translate, text is blured when a scroll indicator appear (not sure if it's visible on the picture, but for me it was clearly visible on my screen)

image

pascaldelange (approved):

PR #532, opened 2024-09-02
custom list repository

Pascal-Delange:
Goes with https://github.com/checkmarble/marble-backend/pull/663: deprecate camel case fields of custom list dtos

- add custom list repository
- snake case dtos:
    - custom lists
    - scenario iteration rules

### DIFF
--- a/packages/app-builder/src/models/custom-list.ts
+++ b/packages/app-builder/src/models/custom-list.ts
@@ -1,0 +1,78 @@
+import {
+  type CreateCustomListBody as CreateCustomListBodyDto,
+  type CustomList as CustomListDto,
+  type CustomListValue as CustomListValueDto,
+  type CustomListWithValues as CustomListWithValuesDto,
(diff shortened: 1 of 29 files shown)


packages/app-builder/src/models/custom-list.ts
```diff
@@ -0,0 +1,78 @@
+import {
+  type CreateCustomListBody as CreateCustomListBodyDto,
+  type CustomList as CustomListDto,
+  type CustomListValue as CustomListValueDto,
+  type CustomListWithValues as CustomListWithValuesDto,
+  type UpdateCustomListBody as UpdateCustomListBodyDto,
```
balzdur, line 6: You can directly rename those type in the API, so they stick to the backend name (`dto.Type`) and give an easy way to find non-refactored dto types from the OAS file.

In short, you need to rename the type in the OAS, generate the API and you're good to go

Pascal-Delange: 👍  makes sense

PR #533, opened 2024-09-06
Thomas/filter decision by inbox

balzdur:
Add future inbox filter on decision list

### DIFF
--- a/packages/marble-api/openapis/marblecore-api.yaml
+++ b/packages/marble-api/openapis/marblecore-api.yaml
@@ -143,6 +143,15 @@ paths:
             items:
               type: string
               format: uuid
+        - name: inbox_id[]
+          description: inbox IDs used to filter the list
+          in: query
+          required: false
+          schema:
+            type: array
+            items:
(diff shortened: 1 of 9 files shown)


packages/marble-api/openapis/marblecore-api.yaml
```diff
@@ -143,6 +143,15 @@ paths:
             items:
               type: string
               format: uuid
+        - name: inbox_id[]
```
Pascal-Delange, line 146: I'm wondering if I don't want to call it `case_inbox_id`, to clarify that it's not a field from the decision itself 🤔

Pascal-Delange (approved):
look good 👍 
just see the one comment about param naming

packages/marble-api/openapis/marblecore-api.yaml
balzdur: As you wish, I will wait for your backend PR before merging this, so I can adapt changes + manually test e2e the feature

PR #537, opened 2024-09-10
new case event: decision reviewed

Pascal-Delange:
Goes with https://github.com/checkmarble/marble-backend/pull/673

I did the necessary changes to display the new kind of event, without touching anything about the new outcome.

image


packages/marble-api/openapis/marblecore-api.yaml
```diff
@@ -4274,6 +4275,40 @@ components:
...
+              type: string
+              format: uuid
+              description: The previous review_status of the decision
+            resource_id:
+              type: string
+              format: uuid
+              description: The id of the decision being reviewed
+            resource_type:
+              type: string
+              description: Must take the value 'decision'
```
balzdur, line 4308: If it will always be `'decision'` in that case, you can improve this by doing 
```suggestion
            resource_type:
              type: string
              enum: ['decision']
```

balzdur (approved):

packages/marble-api/openapis/marblecore-api.yaml
Pascal-Delange: Right 👍 
It's a bit awkward to do an enum with just one option, but I guess that's a side effect of how we split the event into different possible options in the openapi, and it serves to document the values just as well !

PR #544, opened 2024-09-18
Thomas/mar 638 filter decisions by scheduled exec

balzdur:
In this PR :
- implement the decision filter on scheduled executions
  - lazilly fetch the list of existing scheduled executions
  - filter on Scenario name + startedAt (to be discussed)
- add a link from the existing `/scheduled-execution` page in order to prepare for "Download" deprecation

**New filter UI:**
output

**Link from `/scheduled-execution` page:**
output


packages/app-builder/src/components/Decisions/Filters/FilterDetail/ScheduledExecutionFilter.tsx
```diff
@@ -0,0 +1,124 @@
...
+    if (loadFetcher.state === 'idle' && !loadFetcher.data) {
+      loadFetcher.load(
+        getRoute('/ressources/decisions/list-scheduled-execution'),
+      );
+    }
+  }, [loadFetcher]);
+  const successfullScheduledExecutions = React.useMemo(
+    () =>
+      loadFetcher.data
+        ?.filter(({ status }) => status === 'success')
```
Pascal-Delange, line 30: This is valid today but is probably no longer going to be pretty soon, in fact as soon as I start making scheduled execs execute piece by piece instead of in one big transaction.
So: works right now, but we might as well not have this logic and wording implemented to begin with.

Pascal-Delange (approved):

packages/app-builder/src/components/Decisions/Filters/FilterDetail/ScheduledExecutionFilter.tsx
balzdur: If there is still a valid condition to filter all scheduled executions we want to display, it's ok

PR #554, opened 2024-10-01
feat(i18n): improve lng persistence

balzdur:
Small refacto to store the `lng` preference in a separate cookie.
Before that, the `lng` pref was stored on the user session. On each logout, the user session is destroyed and thus the `lng` pref is lost, defaulting to the inferred lng from the browser.


Pascal-Delange (approved):
nice 👍 
I was wondering too, but didn't take the time to ask

PR #555, opened 2024-10-01
chore(deps): move remix-flat-routes to devDependencies

balzdur:
I think I found the issue with the esbuild thing in Docker image


Pascal-Delange (approved):
nice ! 
what does the "deploy" flag do ?

balzdur:
> nice ! 
> 
> what does the "deploy" flag do ?

This should done nothing in that particular case (= position change) it's just to harmonize the CLI usage.

In short, deploy is supposed to create a dedicated node_modules with the required deps for a package inside a monorepo

PR #556, opened 2024-10-03
more granular display of batch job status & output following backend changes

Pascal-Delange:
See https://github.com/checkmarble/marble-backend/pull/690
For now, the number of created decisions is only written at the end of the execution, I'll try later to partially update it at checkpoints during the execution.

I now expose a "partial failure" status in the back-end, but I assumed for now that from a customer point of view it should still be displayed as, and called, "failure". But we can iterate on this later.

"Decisions planned" is nullable (and remains null until a list of target object_ids has been extracted during the batch execution). It remains null for old entries. "Decisions created" now is 0 by default.

Capture d’écran 2024-10-03 à 16 20 48

----

This PR also removes the "download decisions" button.


packages/app-builder/public/locales/fr/cases.json
```diff
@@ -7,8 +7,6 @@
   "case.file.added_date": "Ajouté le",
   "case.file.download": "Télécharger",
   "case.file.downloading": "Téléchargement...",
-  "case.file.errors.downloading_decisions_link.auth_error": "Une erreur d'authentification s'est produite lors de la génération du lien de téléchargement. \nReconnectez vous avant de réessayer.",
-  "case.file.errors.downloading_decisions_link.unknown": "Une erreur inconnue s'est produite lors de la génération du lien de téléchargement. \nVeuillez réessayer plus tard.",
```
balzdur, line 11: Probably an error, you need them as in the en trad

balzdur (approved):

packages/app-builder/public/locales/fr/cases.json
Pascal-Delange: indeed, my bad

PR #557, opened 2024-10-09
Thomas/mar 438 factorize the editor combobox

balzdur:
Disclaimer, this is not yet finished but the set of changes is already massive....

In this PR :
- remove the EditableAstNode classes
  - split internal logic in separate methods inside ast-node services
  - only depends on plain JS objects on relying components
  - wrap data derivation from `AstNode` inside reusable hooks
- clean the editor view model
  - create a dedicated `AstNodeViewModel` since we still need the Tree with nodeId and parent connection for now
  - remove some useless adapters in favour of type narrowing (= instead of a "try to adapt and if success do..." juste use TS type guards, it remove a layer of transformation that is pointless IMO)
  - move validation logic in a separate service (under the already existing `validation` folder)
    - factorize business logic to derive validation state (error messages, validation status...)

Modal edits are on purpose removed in this PR, to reach a "usable" builder. It will be handled in the next PR (atm, we can select a "Function" and it will create the default empty node for this function, without the possibility to fix nested errors since the modal is not yet connected).


packages/app-builder/src/models/ast-node-view-model.ts
```diff
@@ -0,0 +1,443 @@
...
+  | AggregationAstNodeViewModel
+  | TimeAddAstNodeViewModel
+  | TimeNowAstNodeViewModel
+  | FuzzyMatchComparatorAstNodeViewModel;
+
+// TODO(EditableAstNode): heavy link to EditableAstNode operandType = 'Function'
+// Function is considered a special type of operand which require modal to edit
+export function isFunctionAstNodeViewModel(
+  node: AstNodeViewModel,
+): node is FunctionAstNodeViewModel {
```
Pascal-Delange, line 348: this here is type narrowing, is that right?

Pascal-Delange (approved):
👍 
I had a cursory read, but perhaps you can run me through the code tomorrow to re-explain (and show) the big changes orally

balzdur:
I'll do a manual review with you side by side next week.
In the meantime, I pushed preparatory work to make infite modals works easily. You can look at the 4th last commits (1 of 2 last if you really want to concentrate on the main idea).

In short :
- only use a "single" modal to handle every `EditableAstNode` (= nodes with a complex edit flow)
- move the modal to the `OperandEditor` comp (so every comp has it's own modal)
- handle the modal content dynamically, based on the selected node.

What we need to do next :
- simplify the `OperandEditor` file by wrapping the state logic inside a dedicated hook
- make `BottomOptions` + `constantCoercision` customisable
- make errors injection inside edit modal work again (it's broken ATM)
- rework the `Aggregation` modal

Pascal-Delange (approved):
Je blind approve, tu me feras le tour dans la semaine :-)

PR #562, opened 2024-10-15
chore: update deps and pnpm

balzdur:
Update deps to solve dependabot issues (see [here](https://github.com/checkmarble/marble-frontend/security/dependabot)).


package.json
```diff
@@ -5,22 +5,27 @@
...
-    "typescript": "5.6.2",
+    "typescript": "5.6.3",
     "vite-tsconfig-paths": "^5.0.1",
-    "vitest": "2.1.1"
+    "vitest": "2.1.3"
+  },
+  "pnpm": {
+    "overrides": {
+      "cookie": "^0.7.2"
+    }
```
balzdur, line 29: I forced `pnpm` to use a specific `cookie` npm pkg version.
Without that, it was using 3 different ones : `0.6.0`, `0.7.1` and `0.7.2`.

it may be due to the fact `pnpm` consider `0.*.*` version different from the classic semver : a minor update is considered "breaking" (so it do not take care of `^` when doing package deps resolution)

But I'm not totatally sure. In the meantime, this should solve our issues (and the forced version is considered non breaking by the pkg manager).

Pascal-Delange (approved):
thanks!

PR #565, opened 2024-10-21
Thomas/blocked and review UI

balzdur:
Implement new Table UI :
- reduce `min-height` of table rows
- change header bg
- remove borders and use alternate colors (white/grey)

More info [on Figma](https://www.figma.com/design/JW6QvnhBtdZDcKvLdg9s5T/Marble-Portal?node-id=11422-44563&node-type=instance&m=dev)

Bonnus: new Block&Review component is also included in this PR

Some examples of the current change:
image
image

Caveat: I must distinguish hovered row frow grey one, so I needed to change the default bg to a stronger grey (see capture bellow) and it reduces the contrast compared to before (contrast between rows / contrast compared to row cells)
image


Pascal-Delange (approved):
👍

PR #568, opened 2024-10-23
Thomas/sample sentry

balzdur:
Sentry config update :
- remove noise errors
- reduce trace sampling
- remove source maps from build (I expect it will reduce the image size 🤞🏽)


packages/app-builder/src/entry.client.tsx
```diff
@@ -45,6 +52,16 @@ Sentry.init({
   replaysOnErrorSampleRate: 1.0,
 });
 
+function isBrowserExtensionError(exception: unknown): boolean {
```
Pascal-Delange, line 55: line 51 above, let's reduce strongly (say to 0.01)

Pascal-Delange (approved):

PR #570, opened 2024-10-24
Thomas/decisions list

balzdur:
End of the work on decision list redesign.

NB: I'm not totally ISO Figma on :
- case UI: this is a link like scenario name, I do not think using a "boxed" UI is usefull (especially with new white/grey rows)
- keep version number but close to the scenario (ISO what is proposed in the case manager revamp on Figma)

Now, we can only navigate to the version by clicking the scenario (not the scenario): this should be mostly the same since we do not yet have the scenario home page

image

### DIFF
--- a/packages/app-builder/src/utils/format.ts
+++ b/packages/app-builder/src/utils/format.ts
@@ -11,16 +11,22 @@ import { Temporal } from 'temporal-polyfill';
 /**
  * Get the language of the user's browser.
  *
- * This is a workaround for the fact that we only support en, and we want to format dates in the user's language.
+ * This is a workaround for the fact we do not allow the user to change the data format.
  * Since we do not store the user's language preferences, we use the browser's language.
  *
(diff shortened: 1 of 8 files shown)


packages/app-builder/src/utils/format.ts
```diff
@@ -11,16 +11,22 @@ import { Temporal } from 'temporal-polyfill';
 /**
  * Get the language of the user's browser.
  *
- * This is a workaround for the fact that we only support en, and we want to format dates in the user's language.
+ * This is a workaround for the fact we do not allow the user to change the data format.
  * Since we do not store the user's language preferences, we use the browser's language.
  *
+ * Note: Prefered translation language and data format language are not mandatory the same.
```
Pascal-Delange, line 17:
```suggestion
 * Note: Prefered translation language and data format language are not necessarily the same.
```

Pascal-Delange (approved):

PR #572, opened 2024-10-25
fix(TriggerObjectDetail): handle long text in trigger object detail

balzdur:
Improve `TriggerObjectDetail`:
- Handle long text with `break-all`
- cleaner UI using grid (ISO Figma) so it's clearer to read

**Before**
image

**After**
image

### DIFF
--- a/packages/app-builder/src/components/Decisions/TriggerObjectDetail.tsx
+++ b/packages/app-builder/src/components/Decisions/TriggerObjectDetail.tsx
@@ -1,6 +1,6 @@
 import { useFormatLanguage } from '@app-builder/utils/format';
 import { parseUnknownData } from '@app-builder/utils/parse';
-import { useMemo } from 'react';
+import * as React from 'react';
 import { useTranslation } from 'react-i18next';
 import * as R from 'remeda';
 import { Collapsible } from 'ui-design-system';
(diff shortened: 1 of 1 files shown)


packages/app-builder/src/components/Decisions/TriggerObjectDetail.tsx
```diff
@@ -27,13 +27,12 @@ export const TriggerObjectDetail = ({
         {t('decisions:trigger_object.type')}
       </Collapsible.Title>
       <Collapsible.Content>
-        <div className="flex flex-col gap-2">
+        <div className="grid grid-cols-[max-content_1fr] gap-2 break-all">
           {parsedTriggerObject.map(([property, data]) => (
-            <div key={property}>
-              <span className="font-semibold capitalize">{property}:</span>
-              &nbsp;
+            <React.Fragment key={property}>
```
Pascal-Delange, line 32: what is the difference in practice ?

Pascal-Delange (approved):

packages/app-builder/src/components/Decisions/TriggerObjectDetail.tsx
balzdur: To clarify :
- `<div>` is a "real" tag element in the DOM
- `<React.Fragment/>` is a tag element in the virtual DOM but do not create a real tag element in the DOM 

Here, I need to move to a `grid` css layout. Grid do not require "rows" to be represented by intermediate `<div>` element (you directly pass an array of cells). So I need to remove the `<div />` to make it works (if you let divs, it will assume each div is a cell).
Since I do a map, I need to put keys on each element (React requirement), to this end, I use a Framgent.


NB: `<> ... </>` is a suggar notation for `<React.Fragment> ... </React.Fragment>` but you can't add attributes (=props) to the suggar notation

PR #573, opened 2024-10-25
chore(ci): update google-github-actions to v2

balzdur:
In this PR :
- inject sentry env vars (since they are not created anymore by the previous dedicated action)
- update some actions to resolve warnings from the CI (due to Node version updation in Github Actions)

### DIFF
--- a/.github/workflows/build_and_deploy.yaml
+++ b/.github/workflows/build_and_deploy.yaml
@@ -55,19 +55,19 @@ jobs:
           tags: ${{ env.IMAGE }}
           file: ./Dockerfile
           build-args: |
-            "SENTRY_ORG=${{ env.SENTRY_ORG }}"
-            "SENTRY_PROJECT=${{ env.SENTRY_PROJECT }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_deploy.yaml
```diff
@@ -55,19 +55,19 @@ jobs:
...
-        uses: 'google-github-actions/auth@v1'
+        uses: 'google-github-actions/auth@v2'
         with:
           credentials_json: '${{ secrets.GCP_SA_KEY }}'
 
       - name: 'Set up Cloud SDK'
-        uses: 'google-github-actions/setup-gcloud@v1'
+        uses: 'google-github-actions/setup-gcloud@v2'
         with:
           install_components: beta
```
balzdur, line 72: Not sure we still need those but this is out of scope

Pascal-Delange: yep I think we do, it's what's used in the next step

Pascal-Delange (approved):

PR #575, opened 2024-10-25
fix(ci): update sentry release name

balzdur:
Last of the week and also (🤞🏽) last of all...

The sentry release name logic do not work for staging (I expect it to work for prod since we do use the tag, but I prefer to use an explicit easy to understand logic)

### DIFF
--- a/.github/workflows/build_and_deploy.yaml
+++ b/.github/workflows/build_and_deploy.yaml
@@ -57,7 +57,7 @@ jobs:
           build-args: |
             SENTRY_ORG=checkmarble
             SENTRY_PROJECT=marble-frontend
-            VERSION=${{ env.VERSION }}
+            SENTRY_RELEASE=${{ env.GITHUB_SHA }}
           secrets: |
             SENTRY_AUTH_TOKEN=${{ secrets.SENTRY_AUTH_TOKEN }}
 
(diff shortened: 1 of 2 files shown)


packages/app-builder/vite.config.ts
```diff
@@ -43,18 +43,11 @@ if (!isVitest) {
   );
 }
 if (isSentryConfigured) {
-  const version = process.env['VERSION'];
-  const name =
-    version && /v\d+\.\d+\.\d+/.test(version)
-      ? version
-      : process.env['COMMIT_SHA'];
-
```
balzdur, line 51: Remove the "smart" logic from the config

.github/workflows/build_and_deploy.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
           build-args: |
             SENTRY_ORG=checkmarble
             SENTRY_PROJECT=marble-frontend
-            VERSION=${{ env.VERSION }}
+            SENTRY_RELEASE=${{ env.GITHUB_SHA }}
```
balzdur, line 60: Use a default, always valid and explicit name: the commit of the release.

Initially, I tried to use the tag for prod release (so it helps when it comes to read Sentry error reports) but I'm exausted with CI fixes... It's good enough like that

Pascal-Delange (approved):

PR #577, opened 2024-10-28
fix(ci): inject SENTRY_RELEASE into Dockerfile

balzdur:
Everyhting work on Github Action... but I forgot to use the env var inside the Dockerfile


Pascal-Delange (approved):
😅

PR #578, opened 2024-10-28
feat(i18n): add arabic translation

morghim:
I add Arabic language with translate and change some i81n files


Pascal-Delange:
Thanks for your PR ! 
I'm taking your word on the wordings, we just need to fix a little something on the layout because it selecting "ar" as a locale seems to invert much of the app's layout left-right directions.
I'll keep you posted.

balzdur (approved):

PR #586, opened 2024-11-06
Thomas/add arabic language

balzdur:
In this PR :
- add missing arabic translations
- improve CSS on the whole app to handle `rtl` and `ltr`
- rename some components to be direction agnostic (naming follow the flexbox convention)

Bonnus: I removed a lot of `<ScrollArea />` usage in the app, something I want to deprecate so it's a win/win


balzdur:
@morghim, if you have time, it would be nice to look at the added translations (only). They are all auto-generated.

In the future, I propose you open an issue with any "weird" UI / bad translation you find in the product, so we can iterate to ensure a godd user experience

Pascal-Delange (commented):
ok,
To be really honest, I'm a bit conflicted about the the whole RightPanel=>EndPanel etc renaming.
It really sounds like too much of an abstraction for the typical developer who'll work on it and read the code (in other words: I don't think anyone will ever call the thing an "end panel" in natural language, so the component also shouldn't be named like that)
I understand "right" is no longer correct in the context of arab/hebrew, but I think it's a fair tradeoff that left/right is inverted in those contexts.

All good for the ltr/rtl flags however.

morghim (approved):
@balzdur i reviewed all added translations everything is good

balzdur:
I reverted to `Right` / `Left` for component renaming + moved to `rtl:` `ltr:` selector to end/start logical property to reduce the CSS boilerplate (source [here](https://web.dev/learn/design/internationalization?hl=fr))

packages/app-builder/src/components/Decisions/DecisionEndPanel.tsx
```diff
@@ -95,26 +94,19 @@ function DecisionRightPanelContent() {
...
+        </span>
+        <RightPanel.Close />
+      </RightPanel.Title>
+      <AddToCase />
     </RightPanel.Content>
   );
 }
 
-export const DecisionRightPanel = {
+export const DecisionRightPannel = {
```
Pascal-Delange, line 109:
```suggestion
export const DecisionRightPanel = {
```

Pascal-Delange (approved):
thanks 🙏

packages/app-builder/src/components/Decisions/DecisionEndPanel.tsx
balzdur: I'll do it in another PR

PR #595, opened 2024-11-15
light error border on NULL values in decision rule execution viewer

Pascal-Delange:
### Dependency
Paired with https://github.com/checkmarble/marble-backend/pull/742

### Cosmetic changes
The first commit renames "path" to "treePath", in all of the places where we pass a "path" (meaning a path to a node in the AST) as props in the builder. This is not directly relevant to this PR (sorry), but I tried to understand how this works and how it's used, and the simple "path" just had too much collision with other types of "path" (namely url path and "database access operator path").
For changes that are relevant with respect to the title of the PR, focus on commits >= 2.

### Next step
`isEmpty` operator should follow, but in another PR.

### Some screenshots

light red on null values with values - null OR true returns true
Capture d’écran 2024-11-15 à 11 45 59

light red on null values with values - null AND true returns error
Capture d’écran 2024-11-15 à 11 45 03

light red on null values with values - null AND false returns false
Capture d’écran 2024-11-15 à 11 44 39

light red on null values in several places
Capture d’écran 2024-11-15 à 11 44 14
Capture d’écran 2024-11-15 à 11 44 09

### in a follow-up commit
light red on field where division by zero occurs
Capture d’écran 2024-11-18 à 17 27 46

### Edge cases
the implementation seems robust to "unusual" AST rules, not constructed with the marble front end (more nesting, etc):
Capture d’écran 2024-11-18 à 21 37 56
Capture d’écran 2024-11-18 à 21 37 49


packages/app-builder/src/services/editor/ast-editor.tsx
```diff
@@ -315,16 +317,56 @@ export function useRootOrAndChildValidation(stringPath: string) {
...
+  const pathLastPart = path[path.length - 1];
+  if (!pathLastPart) {
+    return false;
+  }
+  if (pathLastPart.type !== 'children') {
+    return false;
+  }
+
+  return (
+    parentNode?.name === '/' &&
```
Pascal-Delange, line 368: So, I tried to see if I could add a unique declaration of the known operators somewhere with aliased names, but I quickly realized that this would be too many places to modify code... possible refactoring better left for another PR.

packages/app-builder/src/services/editor/ast-editor.tsx
```diff
@@ -315,16 +317,56 @@ export function useRootOrAndChildValidation(stringPath: string) {
   ]);
 }
 
-export function useValidationStatus(stringPath: string) {
+export function useValidationStatus(
```
Pascal-Delange, line 320: Honestly I'm not super proud of my code here, but this is precisely part of what I'd like to look at together on Wed.
In the meantime, it works and I don't think it's fundamentally broken.

packages/app-builder/src/services/editor/ast-editor.tsx
```diff
@@ -315,16 +317,56 @@ export function useRootOrAndChildValidation(stringPath: string) {
...
+    parentPath?.childPathSegment,
+    valueIsNull,
+  ]);
+}
+
+function isDivisionDenominator(
+  parentNode: Tree<AstNode> | undefined,
+  path: Path,
+) {
+  if (path.length === 0) {
```
Pascal-Delange, line 356: or would I be correct to say that this part is redundant with line 359 below ? (My brain just has difficulty to write indexes that possibly don't exist 🙈 )

packages/app-builder/src/components/Scenario/AstBuilder/RootAstBuilderNode/RootOrWithAnd.tsx
```diff
@@ -181,20 +185,25 @@ function AndOperand({
...
         />
       </div>
     );
-  } else if (displayReturnValues && childBooleanReturnValue) {
+  } else if (
+    displayReturnValues &&
+    childBooleanReturnValue &&
+    !errorMessages.length
+  ) {
+    const { value } = childBooleanReturnValue;
```
balzdur, line 198: I think, this should be improved :
1. use a condition independant from the "return value"
2. adapt the "return value" and return a comp conditionnaly

Something like this 
```suggestion
  } else if (
    displayReturnValues &&
    !errorMessages.length &&
    evaluation?.returnValue?.isOmitted === false
  ) {
    const value = evaluation.returnValue.value;
```

If you really want to keep the "adapter", I would prefer to split it to prevent the `undefined` check (that is IMO a bad convention). Something like :
```ts
type NonOmittedReturnValue = {
      value: ConstantType;
      isOmitted: false;
    }

function hasReturnValue(returnValue: ReturnValue): returnValue is NonOmittedReturnValue { ...}


function adaptReturnValue(returnValue: NonOmittedReturnValue) ....
```

NB: this an example of anormally cluttered code due to what you pointed out on Notion : the mix of view/edit inside the same components

packages/app-builder/src/services/editor/ast-editor.tsx
balzdur: Ok, let's discuss this on wed.

To keep track of it (depending on if we keep the path access or not):

You can push your good renaming proposition further :
1. rename path -> treePath
2. rename `usePath` -> `useTreePath`
3. rename `useParentPath` -> `useParentTreePath`
4. rename `stringPath`-> `stringTreePath` ?

packages/app-builder/src/services/editor/ast-editor.tsx
balzdur: Yes, you can ommit it safelly. Index access return `undefined` (and do not throw) when this is out of bound in JS.

But you can keep this, it's not a problem

balzdur: There is a "new" [at](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Array/at) array accessor that make code like this easier to read IMO (because it's not using the access syntax used in a lot of programming language w\ the same conventions)

packages/app-builder/src/services/editor/ast-editor.tsx
balzdur: I don't see what you mean here. But this perfectly fine to me to use it like this here.

`/` is a value used in our serialization model, this won't change in the future, we do not need to create a variable to represent it IMO.

packages/app-builder/src/services/editor/ast-editor.tsx
```diff
@@ -315,16 +317,56 @@ export function useRootOrAndChildValidation(stringPath: string) {
...
+  returnValue?: ReturnValue,
+) {
   const path = usePath(stringPath);
   const evaluationErrors = useEvaluationErrors(stringPath);
   const parentPath = useParentPath(path);
+  const parentNode = useParentAstNode(parentPath);
+  const isDivByZeroField =
+    isDivisionDenominator(parentNode, path) &&
+    !returnValue?.isOmitted &&
+    returnValue?.value === 0;
```
balzdur, line 331: It will be computed for each render cycle, on each node of the tree (since there is a high probability this is used in a recursive comp structure).

Still, since this return a primitive value, it won't make the above `useMemo` recompute.

packages/app-builder/src/services/editor/ast-editor.tsx
```diff
@@ -315,16 +317,56 @@ export function useRootOrAndChildValidation(stringPath: string) {
...
+    return getValidationStatus({
       evaluationErrors,
+      valueIsNull,
+      isDivByZeroField,
       parentEvaluationErrors,
-      parentPath?.childPathSegment,
-    );
-  }, [evaluationErrors, parentEvaluationErrors, parentPath]);
+      pathSegment: parentPath?.childPathSegment,
+    });
```
balzdur, line 342: I'm not a super fan of this. If possible, I would prefer to have a logic that create an "artificial" `evaluationError`, so we keep the same logic for `getValidationStatus`.

In the end, a little bit like the `"FUNCTION_ERROR"` implementation

packages/marble-api/openapis/marblecore-api.yaml
```diff
@@ -3798,6 +3798,7 @@ components:
       properties:
         code:
           type: integer
+          enum: [100, 200]
```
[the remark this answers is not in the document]
Pascal-Delange: Ok, in fact the codes other than 100 and 200 should no longer come up from the backend - 201 and 202 in particular were different flavors of "when a null-like value may be read", and 0 AKA NoError should mean no nested error object.
That being said, I think it's fair to have the enum value only handled at the model level, so I can do that.

packages/app-builder/src/components/Scenario/AstBuilder/RootAstBuilderNode/RootOrWithAnd.tsx
Pascal-Delange: alright 👍

packages/app-builder/src/models/decision.ts
```diff
@@ -103,9 +103,15 @@ export function isRuleExecutionHit(
...
 
+export enum RuleExecutionErrorCode {
+  DivisionByZero = 100,
+  NullValueFound = 200,
+}
+
 export interface RuleExecutionError extends RuleExecutionCore {
   outcome: 'error';
   error: Error;
+  errorCode: RuleExecutionErrorCode;
```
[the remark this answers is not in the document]
Pascal-Delange: alright, I'm comfortable with this 👍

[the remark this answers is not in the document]
Pascal-Delange: TBH I followed copilot chat suggestions by using the enum, but I see that it goes counter to the existing convention in our front

packages/app-builder/src/services/editor/ast-editor.tsx
Pascal-Delange: It's fine and I'm not going to put too much effort into this right now, but it's a bit close to a "magic string" situation.
I wouldn't really create a _variable_ to represent it either, but a well named constant, may make sense.

packages/app-builder/src/services/editor/ast-editor.tsx
Pascal-Delange: I see, though I was under the impression that this should work fine - no heavy computation involved.
Would you memoize it ?

packages/app-builder/src/services/editor/ast-editor.tsx
Pascal-Delange: Ok, so basically move logic out of getValidationStatus and modify the evaluationErrors instead ?
Get it, that makes sense to me

Pascal-Delange: Ok I pushed some changes to the PR - those that are related to this comment are in a separate commit from the other ones (see "PR suggestions 2" below)

Pascal-Delange: Should still note, however: currently it's only "parent errors" that lead to a "light-error" coloring, however I'm not really sure where that can actually happen anywhere : I can't see this in any case in the rule builder - perhaps it's a "view only" case ?

packages/app-builder/src/services/validation/ast-node-validation.ts
```diff
@@ -146,30 +146,44 @@ export function findArgumentErrorsFromParent(
...
+
+export function getValidationStatus({
+  evaluationErrors,
+  indirectEvaluationErrors,
+}: {
+  evaluationErrors: EvaluationError[];
+  indirectEvaluationErrors: EvaluationError[];
+}): ValidationStatus {
+  if (evaluationErrors.length > 0) return 'error';
+  if (indirectEvaluationErrors.length > 0) return 'light-error';
```
balzdur, line 186: That's not exactly what I have in mind.

The idea is to capitalize on the existing error structure :
1. direct error
2. argument error (from parent)
3. valid

With the `"FUNCTION_ERROR"` example, we modify the returned error tree from the backend validation to add specifc errors to the frontend (based on the error structure)

Here, the idea would be the same, build upfront the error tree using this logic. 
Based on the current implementation, it may be an argument error added to the parent.
Probably something like this :
1. I have a `division_by_zero` error
2. Traverse the tree to find possible error nodes
3. add an argument error when the error condition is true

balzdur (approved):

thynquest (approved):

packages/app-builder/src/services/validation/ast-node-validation.ts
Pascal-Delange: 👍  as discussed, we merge as it is and review the underlying design tomorrow

PR #602, opened 2024-11-20
WIP: isEmpty operator

Pascal-Delange:
Goes with PR https://github.com/checkmarble/marble-backend/pull/743 on the backend.
Implements the IsEmpty and IsNotEmpty operators, but only on the main AST (not for aggregator filters).
The operator returns true if the value is NULL or empty string, else false.

### Illustration
Starting from a binary operator
Capture d’écran 2024-11-20 à 22 25 54

I can remove the right leg by setting the operator to "is empty"
Capture d’écran 2024-11-20 à 22 26 01

and back again
Capture d’écran 2024-11-20 à 22 26 10

it also works with nesting
Capture d’écran 2024-11-20 à 22 26 25

Decision view display does basically what we can expect it to do
Capture d’écran 2024-11-20 à 22 27 43

### DIFF
--- a/packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/MainAstLine.tsx
+++ b/packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/MainAstLine.tsx
@@ -7,10 +7,6 @@ import {
   NewUndefinedAstNode,
 } from '@app-builder/models';
 import {
-  isBinaryMainAstOperatorFunction,
-  isUnaryMainAstOperatorFunction,
-} from '@app-builder/models/editable-operators';
-import {
   useAstNodeEditorActions,
   useEvaluationErrors,
 } from '@app-builder/services/editor/ast-editor';
@@ -39,8 +35,7 @@ export function MainAstBinaryOperatorLine({
   viewOnly?: boolean;
   root?: boolean;
 }) {
-  const { setAstNodeAtPath, setOperatorAtPath, remove } =
-    useAstNodeEditorActions();
+  const { setAstNodeAtPath, setOperatorAtPath } = useAstNodeEditorActions();
 
   function addNestedChild(stringPath: string, child: AstNode) {
     setAstNodeAtPath(stringPath, NewNestedChild(child));
@@ -79,9 +74,6 @@ export function MainAstBinaryOperatorLine({
           value={mainAstNode.name}
           setValue={(operator: (typeof operators)[number]) => {
             setOperatorAtPath(treePath, operator);
-            if (isUnaryMainAstOperatorFunction(operator)) {
-              remove(rightPath);
-            }
           }}
           validationStatus={evaluationErrors.length > 0 ? 'error' : 'valid'}
           viewOnly={viewOnly}
@@ -121,8 +113,7 @@ export function MainAstUnaryOperatorLine({
   viewOnly?: boolean;
   root?: boolean;
 }) {
-  const { setAstNodeAtPath, setOperatorAtPath, appendChild } =
-    useAstNodeEditorActions();
+  const { setAstNodeAtPath, setOperatorAtPath } = useAstNodeEditorActions();
 
   const operators = useMainAstOperatorFunctions();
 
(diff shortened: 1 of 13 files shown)


packages/marble-api/openapis/marblecore-api.yaml
```diff
@@ -5012,12 +5012,9 @@ components:
       type: object
       required:
         - name
-        - number_of_arguments
```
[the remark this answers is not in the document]
Pascal-Delange: yep, was planning to do that

packages/app-builder/src/models/editable-operators.ts
```diff
@@ -21,25 +21,68 @@ const twoLineOperandOperatorFunctions = [
...
+export function isUnaryMainAstOperatorFunction(
+  value: string,
+): value is UnaryMainAstOperatorFunction {
+  return (unaryMainAstOperatorFunctions as ReadonlyArray<string>).includes(
+    value,
+  );
+}
+
+// define a subset of MainAstOperatorFunction with only binary operators
+const binaryMainAstOperatorFunctions = [
```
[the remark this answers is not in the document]
Pascal-Delange: ok, I tried for 15min, TBH I couldn't get anything reasonable while keeping an ok-looking implem of sortMainAstOperatorFunctions. Typescript keeps getting in the way of reasonable things. 
I have something intermediate where I use separate definitions for binary and unary and a union type for the base case, but I still define the sorted const array for the sake of using in the sort function.

packages/app-builder/src/models/ast-node.ts
```diff
@@ -524,22 +530,60 @@ export function isOrWithAndAstNode(
...
+  if (astNode.name == null || !isMainAstOperatorFunction(astNode.name)) {
+    return false;
+  }
 
   return true;
 }
+
+export function isMainAstUnaryNode(
+  astNode: AstNode,
+): astNode is MainAstUnaryNode {
```
Pascal-Delange, line 571: Typescript question here.
This indication on the return value is never _enforced_ anywhere, right ? It's just something that tells whoever _consumes_ its output that it has this expected form.
Which means that, if the implementation of the function runs against the declared output type, I'm liable to get unexpected prod errors, correct ?

balzdur: Yes.
Typescript blindly accept the type predicate and use it in static type analysys.
The only help you have from TS here is the certainty the implementation is valid TS code, but you can write wrong type predicates
```ts
function isNumber(value: string | number): value is number {
    return typeof value === 'string';
}

// example of help you can get from TS, which is not related to the predicate
function isNumber(value: string | number): value is number {
  return value.prop === 'number'; // Property 'prop' does not exist on type 'string | number'
}

```

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/MainAstLine.tsx
```diff
@@ -100,6 +110,57 @@ export function TwoOperandsLine({
...
+          }}
+          viewOnly={viewOnly}
+        />
+        <Operator
+          value={mainAstNode.name}
+          setValue={(operator: (typeof operators)[number]) => {
+            setOperatorAtPath(treePath, operator);
+            if (isBinaryMainAstOperatorFunction(operator)) {
+              appendChild(treePath, NewUndefinedAstNode());
+            }
```
balzdur, line 152: Same remark here

balzdur (approved):

packages/app-builder/src/models/ast-node.ts
thynquest: Yes if the type narrowing goes wrong you could get prod errors on this one.

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/MainAstLine.tsx
```diff
@@ -69,9 +76,12 @@ export function TwoOperandsLine({
           viewOnly={viewOnly}
         />
         <Operator
-          value={twoLineOperandAstNode.name}
+          value={mainAstNode.name}
           setValue={(operator: (typeof operators)[number]) => {
             setOperatorAtPath(treePath, operator);
+            if (isUnaryMainAstOperatorFunction(operator)) {
+              remove(rightPath);
+            }
```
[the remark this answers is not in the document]
Pascal-Delange: yea sure, I guess that makes sense

PR #605, opened 2024-11-25
WIP: explore d3 library

balzdur:
This is a PoC of D3 to handle the testrun charts display.

In this PR :
- fake `http://localhost:3000/d3` page with hardcoded data
- creation of a D3 chart, rendered with React (not the D3 primitives), that should look like the first graph on Figma.


packages/app-builder/src/routes/_builder+/d3.tsx
```diff
@@ -0,0 +1,251 @@
...
+        version,
+        {
+          outcomeOffset,
+          yScale: d3
+            .scaleLinear()
+            .domain([0, 1])
+            .range([
+              height - (numberOfOutcomes - 1) * gap - margin.bottom,
+              margin.top,
+            ]),
```
balzdur, line 167: This weird computation is required to handle the "gaps" between each part of the stack. This depends on each stack so we need to compute an `yScale` per stack.
image

packages/app-builder/src/routes/_builder+/d3.tsx
```diff
@@ -0,0 +1,251 @@
...
+        </g>
+      ))}
+
+      <use
+        href={`${iconsSVGSpriteHref}#arrow-right`}
+        height={40}
+        width={40}
+        x={(width + margin.left - margin.right - 40) / 2}
+        y={(height + margin.top - margin.bottom - 40) / 2}
+      />
```
balzdur, line 198: Reuse of an existing svg icon (it's the abstracted SVG code hidden inside `Icon` component)

PR #611, opened 2024-11-27
draft: oas for testrun

thynquest:


packages/marble-api/openapis/marblecore-api.yaml
```diff
@@ -3269,6 +3271,126 @@ paths:
...
+      security:
+        - bearerAuth: []
+      parameters:
+        - name: scenarioId
+          description: ID of the scenario that need to be fetched
+          in: query
+          required: true
+          schema:
+            type: string
+            format: uuid
```
balzdur, line 3289: Since we do need to GET testruns by ID, I changed the signature from
```
GET /scenario-testrun/{scenarioId}
```
To
```
GET /scenario-testrun?scenarioId=... // use of a query param
```

Let me know if it's ok.

Another solution is to move everything link to a testrun under something like
```
GET /scenario/{scenarioId}/testrun
POST /scenario/{scenarioId}/testrun
...
```
But it doesn't make sens to fill the `{scenarioId}` on API endpoint using `{testRunId}`

packages/marble-api/openapis/marblecore-api.yaml
```diff
@@ -3269,6 +3271,126 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TestRunDto'
+        '401':
+          $ref: '#/components/responses/401'
+        '403':
+          $ref: '#/components/responses/403'
+        '404':
+          $ref: '#/components/responses/404'
```
balzdur, line 3358: Required for the testrun detail page.
The current endpoint is not supposed to return graph data (the APi for this is out of scope of this PR, existing work in [this openned backend PR](https://github.com/checkmarble/marble-backend/pull/746))

packages/marble-api/openapis/marblecore-api.yaml
```diff
@@ -3269,6 +3271,126 @@ paths:
...
```
balzdur, line 3393: This is the endpoint I'm totally not fan of (any proposition is welcomed).

The ultimate goal is to allow to "down" an ongoing testrun.

Cancelling ongoing testrun looks necessary when you try to create a new one.
Still, this can be handled by the backend usecase, and we can remove this from the API. It won't prevent the frontend to display a warning modal (more or less like the information modal displayed on draft creation)

packages/marble-api/openapis/marblecore-api.yaml
```diff
@@ -5214,6 +5336,75 @@ components:
...
+          type: string
+          format: date-time
+        end_date:
+          type: string
+          format: date-time
+        creator_id:
+          type: string
+          format: uuid
+        status:
+          $ref: '#/components/schemas/TestRunStatusDto'
```
balzdur, line 5376: This is the full list of what the frontend expect (according to the Figma).

Still, one thing is not included: an optionnal "log" info to informa user on the "why" an existing testrun transition to "down" status (ex: Manually cancelled, Cancelled due to live version update on the scenario, cancelled by another testrun....)

From the Figma, this part of the design (choice for now is to make this column out of scope for the first iteration on testrun):
image

packages/marble-api/openapis/marblecore-api.yaml
```diff
@@ -5214,6 +5336,75 @@ components:
...
+          format: uuid
+        phantom_iteration_id:
+          type: string
+          format: uuid
+        start_date:
+          type: string
+          format: date-time
+        end_date:
+          type: string
+          format: date-time
```
balzdur, line 5400: We can discuss on this DTO :
- `ref_iteration_id` seems unecessary with current business rules (= the only accepted version is the current live one) but we can still provide this in the payload and check this is the live_id on the backend (so in a future version, the API is ready to allow to pick a pair of version).
- `start_date` seems unecessary (unwanted)

balzdur (commented):
I merged the work directly inside the `marblecore-api.yaml` file.

I make some changes on the DTO, lot of them require to be reviewed.

packages/marble-api/openapis/marblecore-api.yaml
thynquest: the changed signature is ok for me. for the moment let's keep it simple, we will be able to change of needed. I will adapt the backend accordingly

packages/marble-api/openapis/marblecore-api.yaml
Pascal-Delange: agreed on this - a test run should only be created on the currently live version (and cannot be created if none is live) - so we can remove this from the endpoint.
Similarly, I'd not allow to set a start date in the future - just use "now" (so no need to expose this either).
Finally, I'd go for just "scenario_iteration_id" (or if it's too ambiguous, "test_iteration_id" or such) instead of "phantom_iteration_id" : i find the concept of "phantom decisions" to be a convenient concept and naming to work with on the backend side, but IMO it's an implementation detail and I'd rather avoid "leaking" this naming into the rest of the product.

packages/marble-api/openapis/marblecore-api.yaml
thynquest: let's decide about this later.. I think it is a product design issue; it is not necessary for the v0

packages/marble-api/openapis/marblecore-api.yaml
thynquest: done .. I will integrate it

packages/marble-api/openapis/marblecore-api.yaml
thynquest: works for me.

Pascal-Delange (approved):

PR #612, opened 2024-11-28
Feature: timestamp extract operator

Pascal-Delange:
## Dependencies
Back-end PR https://github.com/checkmarble/marble-backend/pull/757

## Contents
- builder can use the "extract timestamp part" operator function in the rule builder.
- implemented in a modale, the modale contains some inline doc on what to expect as an output value
- new settings page for scenario execution related settings, containing just one entry for now: the timezone to use by default when interpreting timestamps as date/time in a scenario. Only editable for admins.

## Screenshots
- A non-admin user sees a warning when editing the operator if the org setting is not set
Capture d’écran 2024-12-05 à 14 03 07

- but he cannot edit it on the settings page
Capture d’écran 2024-12-05 à 14 03 24

- once the value is set, he can see it
Capture d’écran 2024-12-05 à 14 06 20

---
- An admin user is invited to edit the org's timezone setting when working on the operand. **NB** I wasn't able to make it open the link in a new tab, so you are at risk of losing the edit, but I can live with it since this is a "only once in org's life" action
Capture d’écran 2024-12-05 à 14 03 42

- by default we display "UTC", which is used as a fallback value, but a value is only set when selecting a value from the combobox
Capture d’écran 2024-12-05 à 14 03 50

- Once set, we just display the value here (and you can still update it as an admin)
Capture d’écran 2024-12-05 à 14 04 02

- When the value is set, the warning goes away in the rule editor
Capture d’écran 2024-12-05 à 14 04 18


packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/Operand/OperandEditor/OperandEditModal/TimestampExtract/TimestampField.tsx
```diff
@@ -0,0 +1,74 @@
...
+      // Safeguard; should never happen and does not protect against every possible case (ex: if the astNode is not a Timestamp data type)
+      // Possible upgrade: add a validation on the operand data type (same condition than the one used inside useTimestampFieldOptions)
+      if (!isTimestampFieldAstNode(newSelection)) {
+        toast.error(t('scenarios:edit_date.field_is_not_supported'));
+        return;
+      }
+      onChange(newSelection);
+    },
+    [onChange, t],
+  );
```
balzdur, line 56: I'm not a huge fan of this kind of conditionnal save. By experience, I prefer a structure open to "any kind of modification".
1. Narrowing the possibilities (to prevent errors) can be done using a whitelist of possible values (mostly something like the `coerceToConstant` above already do)
2. In the end, if a "bad" node is saved, this is the validation system that is responsible of informing/helping the user to fix this

IMO, preventing some editting at "save level" is rarelly a good thing.

Still, I must admit that in this perticular case, since we do not support yet "onChange" validation in modals, it could makes sense.

balzdur (approved):

balzdur (approved):

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/Operand/OperandEditor/OperandEditModal/TimestampExtract/TimestampExtract.tsx
```diff
@@ -0,0 +1,240 @@
...
+  currentUser: CurrentUser,
+  t: TFunction<['scenarios']>,
+): React.ReactNode {
+  return isAdmin(currentUser) ? (
+    <span className="text-red-100">
+      <Trans
+        t={t}
+        i18nKey="scenarios:edit_timestamp_extract.missing_default_timezone_admin"
+        components={{
+          DocLink: (
```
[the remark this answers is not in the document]
Pascal-Delange: ok sure. I stuck too close to other translations here I guess.

packages/app-builder/src/routes/_builder+/settings+/scenarios.tsx
```diff
@@ -0,0 +1,65 @@
...
+    authService.isAuthenticated(request, {
+      failureRedirect: getRoute('/sign-in'),
+    }),
+  ]);
+
+  return json({
+    organization: org,
+    user,
+  });
+}
```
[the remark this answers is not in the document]
Pascal-Delange: Aah, yes sorry.
That was a bit dumb and I clearly didn't proof-read myself here.

packages/app-builder/src/components/Scenario/AstBuilder/AstBuilderNode/Operand/OperandEditor/OperandEditModal/TimestampExtract/TimestampField.tsx
Pascal-Delange: Well TBH this file is a pure copy/paste from the file you wrote for the TimeAddEdit modale, as discussed :D 
I have blocking work to finish on my backend PR for Christopher so I'll go ahead and merge this one without reworking the TimestampField component, but point taken and we'll see when we touch this again.

balzdur: NVM, I red to fast.
In that perticular case, since the `onChange` expect a `TimestampFieldAstNode` node, you need to narrow the type (no choice).

This is the only way to do this for now.

PR #613, opened 2024-11-29
test: fix translation type error

Pascal-Delange:
tentative: fix type error
I tried to follow this suggestion, but looks like it failed https://www.notion.so/checkmarble/Internationalisation-14cd1b7bc7c480d2a83cdcaced83e43a?pvs=4#14cd1b7bc7c4805aafebf3a02adc285f


balzdur (approved):
I pushed a new commit + rebased the branch on your feature branch.

In short :
- use of templated string (for a robust i18n support)
- fix formatting errors

You still need to check the translation data. Non exaustive list :
- arabic may be `"{{timestamp}} من {{operator}}"` isntead of `"{{operator}} من {{timestamp}}"` (since it may be inverted in the UI dew to rtl)
- the formatted data may not look like what you want (I spotted some `[...]` while rebasing)

FYI, the error disapeared on my side after by doing this, but I definitely think templated strings should be done using i18next interpolation (not templated hardcoded strings):
```ts
// Before
return `${getOperatorName(context.t, part)} ${context.t('scenarios:edit_timestamp_extract.from')} ${timestampStr}`;

//After
const foo: string = context.t('scenarios:edit_timestamp_extract.from');
return `${getOperatorName(context.t, part)} ${foo} ${timestampStr}`;
```

PR #614, opened 2024-12-02
test: try searchable combobox

Pascal-Delange:
I ran an experiment to use a searchable combobox for this, taking heavy inspiration from the edit pivot ressource.
Two questions:
- it's super super horribly slow when I type. Is there any obvious fix ? (the list of timezones is ~500 strings)
- is this one of the cases where we need react hook form rather than conform ?


balzdur:
image

Pascal-Delange:
Thanks @balzdur !
Last question though: I still find it sluggish to respond to typing (so that if I type normally, characters get lost). Without the "deferred", it seems to respond more quickly and otherwise seems to work as well. Do you see any possible negative side effects of removing the deferred ?

balzdur:
> Thanks @balzdur ! Last question though: I still find it sluggish to respond to typing (so that if I type normally, characters get lost). Without the "deferred", it seems to respond more quickly and otherwise seems to work as well. Do you see any possible negative side effects of removing the deferred ?

Really good catch 👍🏽 
There was a bunch of places in the app that wrongly used `useDeferredValue`. In short :
- `useDeferredValue` deferr heavy UI update of parts of the UI relying on it (the goal is to keep a good interaction on parts that matters)
- some inputs were wrongly controlled using the deferred value

Pascal-Delange:
So recap for future reference: highlight and matches are computed from the deferred value, but the value in the "input" part of the component should not

PR #617, opened 2024-12-05
fix pagination timeouts

ChibiBlasphem:
Changing how pagination behave with no `totalCount` returned from the backend.

### DIFF
--- a/packages/app-builder/src/hooks/useCursorPaginatedFetcher.ts
+++ b/packages/app-builder/src/hooks/useCursorPaginatedFetcher.ts
@@ -48,6 +48,12 @@ export const useCursorPaginatedFetcher = <T, D = T>({
     });
   }, [paginationState, submit]);
 
+  const [previousInitialData, setPreviousInitialData] = useState(initialData);
+  if (initialData !== previousInitialData && paginationState.isPristine) {
+    setPreviousInitialData(initialData);
+    setData(initialData);
+  }
+
   if (fetcherData !== previousFetcherData && fetcherData) {
     const transformedData =
       'transform' in opts ? opts.transform(fetcherData) : (fetcherData as D);
@@ -60,7 +66,6 @@ export const useCursorPaginatedFetcher = <T, D = T>({
 
   return {
     data,
-    pagination: paginationState,
     update: setData,
     next,
     previous,
--- a/packages/app-builder/src/hooks/useCursorPaginatedFetcher.ts
+++ b/packages/app-builder/src/hooks/useCursorPaginatedFetcher.ts
@@ -1,0 +1,72 @@
+import { type SerializeFrom } from '@remix-run/node';
+import { useFetcher } from '@remix-run/react';
+import qs from 'qs';
+import { useCallback, useEffect, useState } from 'react';
+
+import { useCursorPagination } from './useCursorPagination';
+
+type BaseUseCursorPaginatedFetcherOptions<D> = {
+  initialData: D;
+  getQueryParams?: (cursor: string | null) => Record<string, unknown>;
+  validateData?: (data: D) => boolean;
+};
+
+type UseCursorPaginatedFetcherOptions<T, D> =
+  BaseUseCursorPaginatedFetcherOptions<D> &
+    (D extends T
+      ? Record<string, never>
+      : {
+          transform: (value: SerializeFrom<T>) => D;
+        });
+
+export const useCursorPaginatedFetcher = <T, D = T>({
+  initialData,
+  getQueryParams,
+  validateData,
+  ...opts
+}: UseCursorPaginatedFetcherOptions<T, D>) => {
+  const [previousFetcherData, setPreviousFetcherData] =
+    useState<SerializeFrom<T> | null>(null);
+
+  const [data, setData] = useState(initialData);
+  const updateData = useCallback((data: D) => {
+    setData(data);
+  }, []);
+
+  const {
+    state: paginationState,
+    next,
+    previous,
+    reset,
+  } = useCursorPagination();
+  const { data: fetcherData, submit } = useFetcher<T>();
+
+  useEffect(() => {
+    if (paginationState.isPristine) {
+      return;
+    }
+
+    const queryParams = getQueryParams?.(paginationState.cursor) ?? {};
+    submit(qs.stringify(queryParams, { skipNulls: true }), {
+      method: 'GET',
+    });
+  }, [paginationState, getQueryParams, submit]);
+
+  if (fetcherData !== previousFetcherData && fetcherData) {
+    const transformedData =
+      'transform' in opts ? opts.transform(fetcherData) : (fetcherData as D);
+
+    if (validateData?.(transformedData) ?? true) {
+      setPreviousFetcherData(fetcherData);
+      setData(transformedData);
+    }
+  }
+
+  return {
+    data,
+    update: updateData,
+    next,
+    previous,
+    reset,
+  };
+};
--- a/packages/app-builder/src/hooks/useCursorPaginatedFetcher.ts
+++ b/packages/app-builder/src/hooks/useCursorPaginatedFetcher.ts
@@ -1,7 +1,7 @@
 import { type SerializeFrom } from '@remix-run/node';
 import { useFetcher } from '@remix-run/react';
 import qs from 'qs';
-import { useCallback, useEffect, useState } from 'react';
+import { useEffect, useState } from 'react';
 
 import { useCursorPagination } from './useCursorPagination';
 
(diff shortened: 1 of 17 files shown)


balzdur (approved):

packages/app-builder/src/routes/_builder+/cases+/inboxes.$inboxId.tsx
```diff
@@ -69,50 +70,119 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
-                      startDate: casesFilters.dateRange.startDate || null,
-                    }
-                  : {
-                      type: 'dynamic',
-                      fromNow: casesFilters.dateRange.fromNow,
-                    }
-                : {},
-              offsetId: pagination?.offsetId || null,
-              next: pagination?.next || null,
-              previous: pagination?.previous || null,
```
Pascal-Delange, line 104: just double checking here: "previous" can now be removed from the backend API ?

Pascal-Delange: (if yes, let's also remove it on the openapi spec as an option)

packages/app-builder/src/routes/_builder+/cases+/inboxes.$inboxId.tsx
```diff
@@ -69,50 +70,119 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
+                }
+              : {
+                  type: 'dynamic',
+                  fromNow: filters.dateRange.fromNow,
+                }
+            : {},
+          ...(paginationState.cursor
+            ? {
+                offsetId: paginationState.cursor,
+                next: true,
```
Pascal-Delange, line 121: and then if "previous" is no longer an option, perhaps we also remove the option (and the need) to pass "next" as a query param

packages/app-builder/src/hooks/useCursorPagination.ts
```diff
@@ -0,0 +1,55 @@
...
+      lastAction: 'next',
+      isPristine: false,
+    }));
+  }, []);
+
+  const previous = useCallback(() => {
+    setState((currentState) => ({
+      ...currentState,
+      cursor: currentState.previousCursors.pop() ?? null,
+      previousCursors: currentState.previousCursors,
```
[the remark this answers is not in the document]
ChibiBlasphem: The entire state is used in deps array and the say is returning a new ref, it's only the `previousCursors` that keep the same ref and the variable itself is never used unless we are in a `setState`

packages/app-builder/src/routes/_builder+/decisions+/_index.tsx
```diff
@@ -87,55 +89,135 @@ export async function loader({ request }: LoaderFunctionArgs) {
...
   } = useLoaderData<typeof loader>();
 
+  const [decisionsData, setDecisionsData] = useState(initialDecisionsData);
+  const { items: decisions, ...pagination } = decisionsData;
+
+  const [filters, setFilters] = useState(initialFilters);
+  if (initialFilters !== filters) {
+    setFilters(initialFilters);
+    setDecisionsData(initialDecisionsData);
+  }
```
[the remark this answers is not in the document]
ChibiBlasphem: When adding/removing a filter those are the initialData that are update with the loader as we're doing a call to `navigate`. To ensure the correct data are displayed and avoid using an `useEffect` (cf: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes) it's better to do it this way.

packages/app-builder/src/routes/_builder+/cases+/inboxes.$inboxId.tsx
ChibiBlasphem: Yes, just discussed it with Pascal, next and previous params will be deleted from the input

packages/app-builder/src/routes/_builder+/settings+/inboxes.$inboxId.tsx
```diff
@@ -28,24 +28,27 @@ export const handle = {
...
 
   const inboxId = fromParams(params, 'inboxId');
 
-  const [{ inbox }, caseList, inboxUserRoles] = await Promise.all([
+  const [{ inbox }, inboxesList, inboxUserRoles] = await Promise.all([
     apiClient.getInbox(inboxId),
-    cases.listCases({ inboxIds: [inboxId] }),
+    inboxApi.listInboxesWithCaseCount(),
     featureAccessService.getInboxUserRoles(),
   ]);
```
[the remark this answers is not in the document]
ChibiBlasphem: Ok I see what you mean after testing it and it was the reason I did both calls. I'm fixing and refactoring that.

packages/app-builder/src/routes/_builder+/decisions+/_index.tsx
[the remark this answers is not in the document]
balzdur: Ok, I think I got it, filters do matter for pagination.
IIUC, this is the implementation of "reset the pagination when filters change". Shouldn't you clear the pagination state too ?

balzdur (approved):

packages/app-builder/src/hooks/useCursorPaginatedFetcher.ts
```diff
@@ -0,0 +1,72 @@
...
+  validateData,
+  ...opts
+}: UseCursorPaginatedFetcherOptions<T, D>) => {
+  const [previousFetcherData, setPreviousFetcherData] =
+    useState<SerializeFrom<T> | null>(null);
+
+  const [data, setData] = useState(initialData);
+  const updateData = useCallback((data: D) => {
+    setData(data);
+  }, []);
```
balzdur, line 34: I think you don't need this, use `setData` directly

packages/app-builder/src/hooks/useCursorPaginatedFetcher.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  useEffect(() => {
+    if (paginationState.isPristine) {
+      return;
+    }
+
+    const queryParams = getQueryParams?.(paginationState.cursor) ?? {};
+    submit(qs.stringify(queryParams, { skipNulls: true }), {
+      method: 'GET',
+    });
```
balzdur, line 52: FYI(not bloquant for now):
Keep in mind in the future you may need to inject `useFetcher` params (ex: if you want to add pagination in a resource, you need to explicitly give the `action` side by side with the `method`)

packages/app-builder/src/routes/_builder+/cases+/inboxes.$inboxId.tsx
```diff
@@ -70,74 +96,30 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
 
-  const [filters, setFilters] = useState(initialFilters);
-  if (initialFilters !== filters) {
-    setFilters(initialFilters);
-    setCasesData(initialCasesData);
+  const [previousInitialCasesData, setPreviousInitialCasesData] =
+    useState(initialCasesData);
+  if (initialCasesData !== previousInitialCasesData) {
+    setPreviousInitialCasesData(initialCasesData);
+    update(initialCasesData);
```
balzdur, line 119: IIUC, this is to sync the data comming from `fetcher` with the internal state (updated by the internal `useFetcher`).

You get the "reset pagination state" covered when filters changed thanks to the `reset()` call above inside `navigateCasesList`.

Test this "side effect" :
1. Open the app, and do some pagination
2. Loose the focus (ex: go to another app or tab)
3. Go back to the app tab (the page loader should have been retriggered)
4. Try to play with pagination (the old state may be persisted here)

My questions :
1. Do we need to reset the state in that case too ? (or am I missing something ?)
2. This looks something we could wrap inside the `useCursorPaginatedFetcher` hook (since the sync looks pretty common to any usage)

balzdur (approved):

balzdur:
Really good job on the abstraction 👏🏽

packages/app-builder/src/hooks/useCursorPaginatedFetcher.ts
ChibiBlasphem: Oh yeah, I don't need it at all, I had other things in there and I didn't notice there was only the setState inside.

packages/app-builder/src/hooks/useCursorPaginatedFetcher.ts
ChibiBlasphem: Ok I take note

packages/app-builder/src/routes/_builder+/cases+/inboxes.$inboxId.tsx
ChibiBlasphem: For the side effect:
I didn't know remix was re-running loaders when switching tabs, I thought it was only done when refreshing the page or navigating the app.
I'll fix this issue

For your questions:
1. Reseting the state is in my sense logical as you modify the filters you do not want to stay on the same pagination as you have a high-chance of modifying the dataset displayed. (The id you've on cursor with current filter may not be available for another filters)
2. Yeah that's what I did initially, but as you can see the URL based navigation and data are managed outside of the scope of the hook and I wanted to keep it that way too for the data update.

PR #618, opened 2024-12-06
Thomas/case manager update

balzdur:
Batch of changes to adress remarks from one of our customer :
- open case detail on `Decisions` tab
- move `History` inside a right panel which is 
  - closed at first (to give more space to see decisions on small screens)
  - displayed on top of the main layout
- display data inside decision detail
  - the trigger object
  - the pivot list

image


Pascal-Delange (approved):
👍

PR #622, opened 2024-12-10
fix: some translation & date time in adapter DTO

carere:


balzdur (approved):

packages/app-builder/src/routes/ressources+/scenarios+/$scenarioId+/testrun+/create.tsx
```diff
@@ -89,10 +91,21 @@ export function CreateTestRun({
...
+          <Button
+            disabled
+            variant="primary"
+            className="isolate h-10 w-fit cursor-not-allowed"
+          >
+            <Icon icon="plus" className="size-6" aria-hidden />
+            {t('scenarios:create_testrun.title')}
+          </Button>
+        </Tooltip.Default>
+      )}
```
balzdur, line 107: Lift up the conditional :
1. If allowed return the modal
2. If not return the disabled button with tooltip only

balzdur, line 107: Lift up the conditional :
1. If allowed return the modal
2. If not return the disabled button with tooltip only

balzdur (approved):

packages/app-builder/src/routes/ressources+/scenarios+/$scenarioId+/testrun+/create.tsx
carere: This is what I'm doing I guess

PR #628, opened 2024-12-12
disable create test run if one is active

Pascal-Delange:


packages/app-builder/src/routes/ressources+/scenarios+/$scenarioId+/testrun+/create.tsx
```diff
@@ -58,18 +61,49 @@ export async function action({ request, params }: ActionFunctionArgs) {
       }),
     );
   } catch (error) {
-    return json(submission.reply());
+    if (isStatusConflictHttpError(error)) {
```
Pascal-Delange, line 64: TBH I did this first, and then I thought to just disable the button. Perhaps we can leave this here anyway, not sure what's best.

carere: Use ternary

carere (approved):

PR #631, opened 2024-12-13
Stopping Current Testrun

carere:


packages/app-builder/src/routes/ressources+/scenarios+/$scenarioId+/testrun+/$testRunId+/cancel.tsx
```diff
@@ -0,0 +1,101 @@
...
+          <div className="flex flex-1 flex-row gap-2">
+            <ModalV2.Close
+              render={<Button className="flex-1" variant="secondary" />}
+            >
+              {t('common:no')}
+            </ModalV2.Close>
+            <Button className="flex-1" variant="primary" type="submit">
+              {t('common:yes')}
+            </Button>
+          </div>
```
balzdur, line 96: **tl;dr**: 
- `"cancel"` is used in all validation modals to let the user revert the change
- `"ACTION_NAME"` is used in the modal trigger, modal title and modal confirmation button

To prevent confusion, we should rename "cancel" testrun for "stop"/"deactivte" (ask in #product chan)

___

I prefer to use "action verb" for validation (so user do not need to read again the modal context to understand what is going on )
```suggestion
          <div className="flex flex-1 flex-row gap-2">
            <ModalV2.Close
              render={<Button className="flex-1" variant="secondary" />}
            >
              {t('common:cancel')}
            </ModalV2.Close>
            <Button className="flex-1" variant="primary" type="submit">
              {t('scenarios:testrun.cancel')}
            </Button>
          </div>
```

And thus, because "cancel" is used for both labels, I would recommand to change "cancel test run" for "stop test run" (and alike, I think you get it)

packages/app-builder/src/routes/_builder+/scenarios+/$scenarioId+/test-run+/$testRunId.tsx
```diff
@@ -75,12 +82,25 @@ export default function TestRun() {
...
           })}
         />
         <p className="line-clamp-2 text-start">{currentScenario.name}</p>
         <TriggerObjectTag>{currentScenario.triggerObjectType}</TriggerObjectTag>
+        <CancelTestRun testRunId={run.id}>
+          <Button
+            variant="secondary"
+            color="red"
+            className="isolate h-10 w-fit"
+            disabled={!hydrated}
```
balzdur, line 96: No need, this will be injected at the modal level

packages/app-builder/src/repositories/TestRunRepository.ts
```diff
@@ -156,14 +157,24 @@ const testrunRuleExecutions: TestRunRuleExecutionCount[] = [
   },
 ];
 
-export const makeGetTestRunRepository2 = () => {
+export const makeGetTestRunRepository = () => {
```
balzdur, line 160: Take care, IIUC, you are connecting the fake repos here (don't merge it)

balzdur (approved):
FYI, if the "cancel testrun" is not required in multiple pages, you could have put it directly on the page (no ressource required) but most of our internal usecase are done with resources so let's keep it like this

packages/app-builder/src/repositories/TestRunRepository.ts
carere: Yes since I need to add the endpoint to the Open Api Spec, I planned on switching back to prod adapter on the next commit

packages/app-builder/src/routes/_builder+/scenarios+/$scenarioId+/test-run+/$testRunId.tsx
carere: What are you talking about ? The `testRunId` prop or the `<Button />` ?

balzdur: The `disabled={!hydrated}`

Pascal-Delange (approved):

PR #641, opened 2024-12-19
Improving Licensing & Feature Access Check

carere:


Pascal-Delange (approved):
I'm tempted to blind approve most parts of this PR, as it's massive with quite some boilerplate. (I read some parts which seemed more impactful)
A little checklist of questions and other input.
- with this, we can remove the license key on the frontend, right ? If so, let's remove the code where it's read
- I remember now that ideally we wanted ideally to mark the paid features for self-hosted customers in "paid" mode (AKA highlight that they're not available in OSS mode) - this is not directly related to this PR, other than in the sense that it's now easy to pilot display changes on the frontend by tweaking the backend endpoint - but let's make a small change to the backend endpoint next week so that it returns at least "test" for everything if the user has the firebase emulator enabled. And while we're at it, ideally mark as "locked" the features that necessitate some setup that isn't done (convoy env variables, etc). The point being: maybe a last good thing in this PR to do is a forth option in feature access (along available, restricted, test) to show a feature that's available in principle but is missing dependencies. Then we can pilot the display from the backend
- FYI, the links for the doc: [rule snoozing](https://docs.checkmarble.com/docs/rule-snoozes), [workflows](https://docs.checkmarble.com/docs/introduction-5), [webhooks](https://docs.checkmarble.com/docs/introduction-3). No doc page exists yet for user roles, test run, sanctions, analytics so let's not display a link (just the "upgrade" link)

PR #643, opened 2025-01-02
refactor: shuffle functions

Pascal-Delange:
### Context
First half of a refactor I wanted to do on the models for ast nodes & operators.
The diff is likely a bit hard to read - try to read it commit by commit if you must.
I'm mostly doing two things here. Before this PR, there were two fairly large models files: `packages/app-builder/src/models/ast-node.ts` and `packages/app-builder/src/models/editable-operators.ts`.

### Things I'm trying to improve
- The former was 700+ lines long, without clear hierarchy, and contained the interfaces and helper functions for all the types of ast nodes we are manipulating. It was just hard to deal with due to its size and lack of structure.
- The latter was annoying me more, because it contained a mix of logic related to actual "ast operators" (nodes in our AST), to "modale operators" (selectors of things like aggregation filters that are not AST nodes), interfaces and helper functions related to the "AST structure" in our builder, and display logic (translations) of all of the above.
- (I also deleted a whole file, whose logic had been refactored elsewhere but where dead code remained, see `packages/app-builder/src/components/Decisions/RulesExecutions/RuleExecutionDetail.tsx`)

### Up next
This PR only moves things around, to try to make the file structure make more sense by separating things that should be separate. This is only half of what I wanted to do, and the other half where I actually change logic will follow in another PR (else it would be really impossible to properly read).
Continued in https://github.com/checkmarble/marble-frontend/pull/646


ChibiBlasphem (approved):
LGTM, that's a great work. I really like how you separated concerns.

PR #645, opened 2025-01-03
fix wrong translations

Pascal-Delange:
The ar and fr translations where incorrect, compared to the en one and what the code expects.


ChibiBlasphem (approved):
LGTM

PR #646, opened 2025-01-03
operator list is now coded in the frontend

Pascal-Delange:
Continues https://github.com/checkmarble/marble-frontend/pull/643, but with logic changes
The main difference is that the list of admissible AST Operators is no longer served from the backend. This was an early "by default" design decision, that did not age well.
Prerequisite for https://github.com/checkmarble/marble-backend/pull/793


packages/app-builder/src/models/operator-functions.ts
ChibiBlasphem, line 1: That's maybe nitpicking but now that `OperatorFunction` has been renamed `OperatorOption` shouldn't we also rename the file?

ChibiBlasphem (approved):
LGTM, I think we still need to simplify how the whole builder is managed in terms of state/context as this PR, even though it is quite small, was a little complex to read and follow.

packages/app-builder/src/models/operator-functions.ts
Pascal-Delange: yep true, I'll rename it

Pascal-Delange:
> LGTM, I think we still need to simplify how the whole builder is managed in terms of state/context as this PR, even though it is quite small, was a little complex to read and follow.

Hard agree. I think there is scope for quite a lot of simplification, the tricky thing is to find where to start. I did this one because it really touched things that I stumbled upon while working on the builder in Nov. I'm open for proposals, whether it's "small touch" refactors (like this, even though it touched a lot of lines of code) or more structural changes (as time allows).

PR #647, opened 2025-01-06
feat: case name templating

ChibiBlasphem:
- Added new StringTemplate AST Node
- Let user update the case name template on workflow page


Pascal-Delange (approved):
This one will have significant (but manageable) conflict with my refactor PR https://github.com/checkmarble/marble-frontend/pull/643, but you go ahead and merge it first.

packages/marble-api/openapis/marblecore-api.yaml
```diff
@@ -4838,8 +4838,9 @@ components:
           type: object
           allOf:
             - $ref: '#/components/schemas/NodeDto'
-        return_type:
+        expected_return_type:
           type: string
+          enum: ['string', 'int', 'float', 'bool', 'datetime']
```
Pascal-Delange, line 4843: just a comment: I prefer "timestamp" to "datetime" here because that is what is actually manipulated in the backend, but this is in no way blocking for this feature so feel free to not make the change now

Pascal-Delange (approved):
👍

packages/marble-api/openapis/marblecore-api.yaml
ChibiBlasphem: i'll remove it from the enum list as for now we do not use it, changing it to timestamp would mean changing the backend and I already merged the PR on this side.

PR #649, opened 2025-01-08
feat: debounce option for the input; useful with large table filtering

alexurdea:


packages/ui-design-system/src/Input/Input.spec.tsx
```diff
@@ -20,4 +20,77 @@ describe('Input', () => {
...
+
+  describe('when debounced prop is used', () => {
+    const user = userEvent.setup({
+      advanceTimers: vi.advanceTimersByTime.bind(vi),
+    });
+
+    beforeEach(() => {
+      vi.useFakeTimers();
+
+      // userEvent.click() fails when used with vi.useFakeTimers()
```
alexurdea, line 32: Actually, all interactions.

Pascal-Delange (commented):
Hi @alexurdea ,
Thanks for your contribution.
On the purely technical side, we already use remeda as a dependency, and I'd rather not add radash.

Regarding the debouncing, it is rather our opinion that debouncing should be the responsibility of the component who imports from the design system - though I'll let @ChibiBlasphem give his opinion if needed.

PR #651, opened 2025-01-13
refactor: change tailwind css colors to fit with figma

ChibiBlasphem:
This PR is divided in 3 parts:
- 1st commit: Add figma colors as new colors suffixed with `figma` to ensure no naming conflict while replacing
- All subsequents commits - to before last: Each commit replace all occurrences of colors with the corresponding figma color
- Last commit: Remove all old colors and figma suffixes


Pascal-Delange (approved):
thanks for taking care of that chore ! 
what's your feeling about all those colors in the div classes and such ? Is it an issue in the long term ?

PR #655, opened 2025-01-15
style: updating the scenario homepage

carere:
This PR update the homepage of the scenario.
As discussed precendently, we now have a section execution with each feature in it's oown block

Screenshot From 2025-01-15 13-20-48


Pascal-Delange (approved):
nice 👍  thanks

PR #663, opened 2025-01-23
Adding BreadCrumb to the app

carere:
- [X] Enable Breadcrumb all around the app


packages/app-builder/src/repositories/TestRunRepository.ts
```diff
@@ -176,8 +177,15 @@ export const makeGetTestRunRepository2 = () => {
       return Promise.resolve();
     },
     listTestRuns: () => Promise.resolve(testruns),
-    listDecisions: () => Promise.resolve(testrunDecisions),
-    listRuleExecutions: () => Promise.resolve(testrunRuleExecutions),
+    listDecisions: async () => {
```
Pascal-Delange, line 180: careful with stubbed repos :-)

packages/app-builder/src/routes/_builder+/api.tsx
```diff
@@ -20,6 +25,18 @@ const SwaggerUI = React.lazy(() => import('swagger-ui-react'));
 
 export const handle = {
   i18n: ['common', 'navigation', 'api'] satisfies Namespace,
+  BreadCrumbs: [
```
Pascal-Delange, line 28: how exactly does this "handle" work ?

Pascal-Delange (changes requested):
nice ! I'll let Christopher give a second opinion on the component :-)
small error to fix on the stubbed repo

packages/app-builder/src/routes/_builder+/api.tsx
carere: The handle is a feature from Remix, it can be defined on each route, then, when you use the hook `useMatches`, you can get all handle from matched routes. That what I use to create the breadcrumb for the app.
More information on the [remix docs](https://remix.run/docs/en/main/guides/breadcrumbs#defining-the-breadcrumbs-for-routes) 😉

packages/app-builder/src/repositories/TestRunRepository.ts
carere: Yes the in-memory repo is not used actually, this modification was used to simulate a timeout a long request to identify a previous bug, so no problem on this one, the real repository is used 😄

Pascal-Delange: you're right, my bad ! I didn't read it correctly

Pascal-Delange (commented):
👍

ChibiBlasphem (approved):
LGTM! Great work

PR #665, opened 2025-01-24
Improve documentation of the env file example 

apognu:
Same rationale as the backend one.

Big difference, though: this actually changes some code to mark some variables as optional. I do not think we use them for anything, from what I can see in the code and the Firebase project, so having them in the configuration file might clutter it for nothing.

### DIFF
--- a/packages/app-builder/src/utils/environment.ts
+++ b/packages/app-builder/src/utils/environment.ts
@@ -27,11 +27,11 @@ const PublicEnvVarsSchema = z.object({
 
   FIREBASE_AUTH_EMULATOR_HOST: z.string().optional(),
   FIREBASE_API_KEY: z.string(),
-  FIREBASE_APP_ID: z.string(),
-  FIREBASE_AUTH_DOMAIN: z.string(),
-  FIREBASE_MESSAGING_SENDER_ID: z.string(),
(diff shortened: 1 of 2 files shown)


packages/app-builder/src/utils/environment.ts
```diff
@@ -27,11 +27,11 @@ const PublicEnvVarsSchema = z.object({
...
   FIREBASE_API_KEY: z.string(),
-  FIREBASE_APP_ID: z.string(),
-  FIREBASE_AUTH_DOMAIN: z.string(),
-  FIREBASE_MESSAGING_SENDER_ID: z.string(),
+  FIREBASE_APP_ID: z.string().optional(),
+  FIREBASE_AUTH_DOMAIN: z.string().optional(),
+  FIREBASE_MESSAGING_SENDER_ID: z.string().optional(),
   FIREBASE_PROJECT_ID: z.string(),
-  FIREBASE_STORAGE_BUCKET: z.string(),
+  FIREBASE_STORAGE_BUCKET: z.string().optional(),
```
Pascal-Delange, line 34: this one I guess can be removed altogether ?

Pascal-Delange (approved):
great !

PR #666, opened 2025-01-27
Evolve Webhook page

carere:


packages/app-builder/src/components/Webhooks/EventTypes.tsx
```diff
@@ -49,7 +52,13 @@ export function FormSelectEvents({
         />
         <FormSelectWithCombobox.ComboboxList>
           {matches.map((event) => (
-            <FormSelectWithCombobox.ComboboxItem key={event} value={event}>
+            <FormSelectWithCombobox.ComboboxItem
+              key={event}
+              value={event}
+              disabled={
+                webhookStatus === 'restricted' ? event.includes('case') : false
```
Pascal-Delange, line 59: I have an opinion here: I think we want explicitly whitelist only the decisions webhooks in this case, rather blacklist everything related to cases (even though with current webhooks coverage it's "equivalent")

packages/app-builder/src/routes/_builder+/settings+/webhooks.tsx
```diff
@@ -138,3 +139,99 @@ export default function Webhooks() {
     </Page.Container>
   );
 }
+
+export function ErrorBoundary() {
```
Pascal-Delange, line 143: This is what the page displays if any error is encountered, is that correct ?

packages/app-builder/src/locales/fr/settings.json
```diff
@@ -109,5 +109,7 @@
   "scenario_default_timezone.more_results_other": "{{count}} fuseaux horaires supplémentaires.\n\nAffinez votre recherche pour les voir.",
   "scenario_default_timezone.not_set": "Non défini",
   "users.role.nudge": "Attribuez des rôles et rationalisez facilement le flux de travail de votre équipe",
-  "webhooks.nudge": "Utilisez des Webhooks pour recevoir des notifications en temps réel chaque fois que des événements se produisent dans votre instance Marble"
+  "webhooks.nudge": "Utilisez des Webhooks pour recevoir des notifications en temps réel chaque fois que des événements se produisent dans votre instance Marble",
+  "webhooks.configuration_error": "Erreur de configuration",
+  "webhooks.convoy_error": "Convoy ne semble pas être correctement configuré sur le backend Marble. \nPour plus d'informations sur les WebHooks, veuillez vous référer à <coclink> la documentation </cocLink>."
```
Pascal-Delange, line 114: Is this actually the correct error message ? In what case exactly do we get this?

carere: We get this error message, when the backend send a 500 when we try to list the available webhooks. It is usually a problem with Convoy not being properly configured.
Unfortunatly I don't have the right Error from the backend, only a opaque 500 Internal Error...

packages/app-builder/src/components/Webhooks/EventTypes.tsx
carere: It was a demand from the stakeholders to display all webhooks, but only disablilng all `case` related webhooks, in order to showcase that those webhooks are only available with premium membership

packages/app-builder/src/locales/fr/settings.json
Pascal-Delange: Ok.
In fact, the error can also be due to the endpoint that is set by the user not responding/having some issues. So not necessarily just a convoy config issue.
But ok, we can live with this message until we perhaps rework the error message passed by the backend in this occasion.

packages/app-builder/src/components/Webhooks/EventTypes.tsx
Pascal-Delange: Yes I understand that, but still: let's just write it like `! event.includes('decision')` rather than `event.includes('case')` so that the condition does not become wrong when we add other webhooks than decisions and cases

carere: I did the modification and I also added some Nudge to signal to users they don't have access to all event types 😉

Pascal-Delange (approved):
nice !

PR #668, opened 2025-01-28
Display nodes that were not executed as skipped.

apognu:
Some nodes can now be skipped during evaluation if their outcome has no way to change the parent's. This displays those as such in the evaluation UI.

image


packages/app-builder/src/components/Scenario/AstBuilder/RootAstBuilderNode/RootOrWithAnd.tsx
```diff
@@ -201,9 +201,19 @@ function AndOperand({
           <Tag
             border="square"
             className="w-full"
-            color={value === null ? 'orange' : value ? 'green' : 'red'}
+            color={
+              evaluation.skipped
+                ? 'grey'
+                : value === null
+                  ? 'green'
```
Pascal-Delange, line 208:
```suggestion
                  ? 'orange'
```

Pascal-Delange (approved):
👍

PR #670, opened 2025-01-28
Pass all feature access to test

carere:


Pascal-Delange (commented):
yea that's the idea, but I'd like to move this logic to the backend so that we always call the back from the front to get the feature accesses :-) (so that it's more ISO to a prod setup)

carere:
So should I remove this part totally ?

Pascal-Delange:
> So should I remove this part totally ?

let me propose a backend PR first, a bit later this week

PR #673, opened 2025-01-28
chore: only execute remix dev tools if 'devTools' query params is present in URL

ChibiBlasphem:


carere (approved):
Nice, so we need to add `?devTools=true` to the url to have access to devtools ?

PR #689, opened 2025-02-13
Add number of datasets selected + React.memo to improve performance

carere:


packages/app-builder/src/routes/_builder+/scenarios+/$scenarioId+/i+/$iterationId+/sanction.tsx
```diff
@@ -150,31 +150,32 @@ export async function action({ request, params }: ActionFunctionArgs) {
...
-    await authService.isAuthenticated(request, {
-      failureRedirect: getRoute('/sign-in'),
-    });
+  const iterationId = fromParams(params, 'iterationId');
+  const formDataDecoded = formDataToObject(formData, {
+    arrays: ['datasets'],
+  });
 
   try {
+    const data = editSanctionFormSchema.parse(formDataDecoded);
```
ChibiBlasphem, line 168: why does it go in the `try` now ? Couldn't you use `safeParse` with and return an object `{ success: false, errors: errors[] }`  or `{ success: true, data }` ?

ChibiBlasphem (commented):
Are all the memo really necessary?

packages/app-builder/src/routes/_builder+/scenarios+/$scenarioId+/i+/$iterationId+/sanction.tsx
carere: I put it in the try block because `parse` throw if it cannot parse, that way I don't have to right the exact same code present in the catch block just before the try-catch ;)

carere:
The `memo` are necessary since in the 2 components where I use the memoized components, React will re-render each component since they all depend on the same value (an array).

packages/app-builder/src/routes/_builder+/scenarios+/$scenarioId+/i+/$iterationId+/sanction.tsx
ChibiBlasphem: Ok, so the failed validation is an unknown error?

carere: And we do not need to return the error since the message shown with the Toast do not include it

ChibiBlasphem (approved):
LGTM

PR #701, opened 2025-02-17
Handling last remark for sanction check config

carere:
- [X] Removing Score Modifier from Sanction check Config
- [X] Save empty sanction when creating new sanction check config
- [X] Prevent double submit when deleting Sanction Check Config
- [X] Set `block_and_review` as default choice for `forced_outcome`
- [X] Enable `rule_group` selector in the sanction check config form
- [X] Back Button on Sanction Check Config redirect to rules page
- [x] Return `undefined` if node is of type `UndefinedNode`
- [x] Clear from `CounterPartyName` nodes which are `UndefinedNode`
- [x] Add zod Validation if we don't have at least `name` or `label` in the query
- [x] Branch Validation for sanction check config

### DIFF
--- a/packages/app-builder/src/routes/_builder+/scenarios+/$scenarioId+/i+/$iterationId+/sanction.tsx
+++ b/packages/app-builder/src/routes/_builder+/scenarios+/$scenarioId+/i+/$iterationId+/sanction.tsx
@@ -133,10 +133,8 @@ const editSanctionFormSchema = z.object({
     z.literal('block_and_review'),
   ]),
   triggerRule: z.any().nullish(),
-  query: z.object({
-    name: z.any(),
-    label: z.any().nullish(),
-  }),
+  counterPartyName: z.any().nullish(),
+  counterPartyLabel: z.any().nullish(),
   counterPartyId: z.any().nullish(),
   datasets: z.array(z.string()),
 });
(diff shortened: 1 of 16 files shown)


packages/app-builder/src/routes/_builder+/scenarios+/$scenarioId+/i+/$iterationId+/sanction.tsx
```diff
@@ -219,44 +221,38 @@ export default function SanctionDetail() {
...
       description: sanctionCheckConfig?.description ?? '',
-      ruleGroup: sanctionCheckConfig?.ruleGroup ?? 'Sanction check',
+      ruleGroup: sanctionCheckConfig?.ruleGroup ?? 'Sanction Check',
       datasets: sanctionCheckConfig?.datasets ?? [],
-      forceOutcome:
-        (sanctionCheckConfig?.forceOutcome as SanctionOutcome) ?? 'decline',
-      scoreModifier: sanctionCheckConfig?.scoreModifier ?? 0,
+      forcedOutcome:
+        (sanctionCheckConfig?.forcedOutcome as SanctionOutcome) ??
+        'block_and_review',
```
Pascal-Delange, line 252: thinking back on this: in fact, since with those changes we now create the config "for real" before we actually visit the page, and the field "forced_outcome" is not actually nullable, I guess it's rather on the backend side that we should use the right default value => I'll do this tomorrow (as well as the error handling when some values are missing)

Pascal-Delange (approved):
allez 💪  good job !

PR #702, opened 2025-02-17
fix: fix sanctions checks bugs

ChibiBlasphem:
- Add information from sanction check in decision detail
- Keep button to sanction check on case when it's reviewed, changed the wording accordingly
- Add back button before breadcrumb leading to case page on sanction pages
- Make the decision able to be approved only if the sanction check (if there's one) has the status "no_hit"


Pascal-Delange (approved):
great, thanks !

PR #710, opened 2025-02-22
chore: change prettier printWidth

ChibiBlasphem:
Change default prettier config for printWidth (80 characters) as it doesn't really suit a React + Typescript codebase style and creates a lot of line-breaks which are hindering the global reading comprehension of the code.


Pascal-Delange (approved):
👍  💯

PR #712, opened 2025-02-25
feat: display app and api version in the Help center

ChibiBlasphem:
Display frontend and backend version in the footer of the Help center:
image


Pascal-Delange (approved):
👍 
no question on the implem, just one about the design: what do we display if we are not on a github version tag? Empty field? The commit hash ?

ChibiBlasphem:
> 👍 no question on the implem, just one about the design: what do we display if we are not on a github version tag? Empty field? The commit hash ?

I committed 2 new edited files which should answer your question. If those variables are not defined, the frontend version is displayed at "unknown"

PR #720, opened 2025-02-28
the env var with app version was not exposed in the final docker layer

Pascal-Delange:


apognu (approved):

apognu:
To be refined later, ideally, we want the var to be baked in at compile time.

PR #726, opened 2025-03-04
feat: display wikidata Id as a link

ChibiBlasphem:
Display wikidata ID as a link in sanction check match:
image


Pascal-Delange (approved):
🙌

PR #724, opened 2025-03-04
feat: make trigger in sanctioncheck optional

ChibiBlasphem:
This PR makes the sanction check rule trigger condition optional:
image

PR #729, opened 2025-03-05
fix: fixed rule builder broken and initial value

ChibiBlasphem:
Side effect of https://github.com/checkmarble/marble-frontend/pull/724 made the rule builder unusable.
This PR fixes this by setting the right default value (which is different wether it's a rule or a sanction trigger) from the outside of the component


Pascal-Delange (commented):
Hmm
Open question, would it not be a good plan to rather make the default value the same for both components ?

ChibiBlasphem:
> Hmm Open question, would it not be a good plan to rather make the default value the same for both components ? (both from backend and frontend point of view)

I don't think so, formerly (before trigger was optional) values were already different:
- Rule is based on Or with ands
- Trigger is a And

What changed is:
- The new rule page was bugged as it was created with an And node (which is why I didn't think it was used on this page)
- Now as the sanction trigger changed to be optional, I just switched the valued from a And node to an undefined one.

carere (approved):
Nice modification :)

PR #732, opened 2025-03-05
Better error messages in aggregation modale

Pascal-Delange:
## Prerequisites
See https://github.com/checkmarble/marble-backend/pull/889 for error codes definition.

## Content 

Display a "pretty" error message below the field selector in the aggregation modale when the state of the node is not completely set, or inconsistent choices are made.

## Screenshots

No field chosen:
Capture d’écran 2025-03-05 à 23 35 35

Field chosen has a type that is not compatible with this aggregation function:
Capture d’écran 2025-03-05 à 23 35 45


ChibiBlasphem (approved):
LGTM

PR #735, opened 2025-03-11
refactor: restructure ast builder MAR-796

ChibiBlasphem:


carere (approved):
Great PR !! Happy to see that our AstBuilder will be way more performant 😄
Just some questions about the comments you let on the base code, do they need to be handled before merging or is it planned to handle them after the merge ?

packages/app-builder/src/components/AstBuilder/edition/node-store.ts
```diff
@@ -0,0 +1,70 @@
...
+    validationFn,
+  }: {
+    initialNode: AstNode;
+    initialEvaluation: FlatNodeEvaluation[];
+    validationFn: AstBuilderValidationFn;
+  }): AstBuilderNodeStoreValue {
+    return { node: initialNode, evaluation: initialEvaluation, copiedNode: null, validationFn };
+  },
+}).withActions({
+  setNodeAtPath(api, path: string, newNode: AstNode) {
```
Pascal-Delange, line 30: just to understand here, what is api? Btw why don't we need to set its type for ts ?

Pascal-Delange (commented):
I read a good part of it, though not all.
Good job !

PR #736, opened 2025-03-12
Little improvement to Case

carere:
- [X] Remove modal when update status of a case
- [x] Open a case from a new tab

### DIFF
--- a/packages/app-builder/src/components/Cases/CasesList.tsx
+++ b/packages/app-builder/src/components/Cases/CasesList.tsx
@@ -35,6 +35,20 @@ export function CasesList({
 
   const columns = useMemo(
     () => [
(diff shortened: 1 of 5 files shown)


Pascal-Delange (commented):
really sorry, the [linear ticket](https://linear.app/marble-eu/issue/MAR-787/remove-the-confirmation-modale-when-updating-case-status) I created was a bit incomplete or misleading, the details were in [the doc attached in the ticket](https://linear.app/marble-eu/document/explorations-on-designs-for-the-case-manager-revamp-30337e913486) : for now at least, we wanted to remove the modale only when the status is changed to "open" or "investigating" 🙏

carere:
@Pascal-Delange you can review it if you want

PR #741, opened 2025-03-18
fix: manage potential errors when calling dataset freshness

ChibiBlasphem:


apognu (approved):

packages/app-builder/src/routes/_builder+/_layout.tsx
```diff
@@ -69,15 +70,21 @@ export async function loader({ request }: LoaderFunctionArgs) {
...
+  ]);
+
+  let datasetFreshnessInfo: { lastExport: string } | null = null;
+  try {
+    datasetFreshnessInfo = await getDatasetFreshnessInfo(sanctionCheck, scenario);
+  } catch (err) {
+    if (!isHttpError(err)) {
+      captureRemixServerException(err, 'remix.server', request, true);
+    }
+  }
```
siiick, line 87: What do you think if we turn the `Promise.all()` into a `Promise.allSettled()` to keep the `getDatasetFreshnessInfo()` method parallelised and handle the errors gracefully ?

PR #744, opened 2025-03-18
Update Dependencies

carere:
The goal here, is to update maximum dependencies, without taking into account the main one:
- React
- Remix
- Sentry
- TailwindCSS


ChibiBlasphem (commented):
I see some dev dependencies changes, like: `oazapfts` or `vitest` did you test running tests or api generation to see if everything is good? (tests should be ok as they're ok on CI)

carere:
@ChibiBlasphem It' s working all generation command (routes + marble-api) are working as exepected)

PR #750, opened 2025-03-24
Add Snooze Filter to Case List

carere:
Simply add a toggle to display or not cases which were snoozed


packages/app-builder/src/components/Cases/Filters/CasesFiltersContext.tsx
```diff
@@ -13,6 +13,7 @@ export const casesFiltersSchema = z.object({
   statuses: z.array(z.enum(caseStatuses)).optional(),
   dateRange: dateRangeSchema.optional(),
   name: z.string().optional(),
+  snoozed: z.union([z.literal('true'), z.literal('false')]).optional(),
```
siiick, line 16: suggestion: Maybe we can use an enum here and transform it to a boolean.
```suggestion
  snoozed: z
    .enum(['true', 'false'])
    .transform((val) => val === 'true')
    .optional(),
```

This way we can setup the CasesFiltersForm.snoozed as an optional boolean.
WDYT ?

packages/app-builder/src/components/Cases/Filters/CasesFiltersContext.tsx
```diff
@@ -29,6 +30,7 @@ export type CasesFiltersForm = {
   statuses: CaseStatus[];
   dateRange: DateRangeFilterForm;
   name?: string;
+  snoozed?: 'true' | 'false';
```
siiick, line 33: Suggestion : Following the previous comment.
```suggestion
  snoozed?: boolean;
```

packages/app-builder/src/routes/_builder+/cases+/inboxes.$inboxId.tsx
```diff
@@ -69,11 +71,19 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
   const filtersForBackend: CaseFilters = {
-    ...parsedQuery.data,
+    ...{
+      ...parsedQuery.data,
+      snoozed:
+        parsedQuery.data.snoozed === undefined
+          ? undefined
+          : parsedQuery.data.snoozed === 'true'
+            ? true
+            : false,
```
siiick, line 85: Not 💯 sure, but if we apply the suggested change in `CasesFiltersForm`, we might be able to remove this, since it’s already an optional boolean.

packages/app-builder/src/components/Cases/Filters/CasesFiltersContext.tsx
carere: The problem here is that the `parseSafeQuery` fucntion decode Search params in query string, and the `qs` library set boolean as `snoozed=true` or `snoozed=false`.
That's why I had to do this gymnastic, else we should use different schema for the same data.
In fact that why I posted my message on Slack, here is a quick win, but we need to look this part of the codebase more thoroughly in order to fix those issues about encoding & decoding query string search params

packages/app-builder/src/routes/_builder+/cases+/inboxes.$inboxId.tsx
carere: Responded on my first answer :)

packages/app-builder/src/components/Cases/Filters/CasesFiltersContext.tsx
carere: Responded on my first answer :)

siiick (approved):
👍

PR #755, opened 2025-03-25
feat: case manager page template

ChibiBlasphem:
- Set case manager page structure in (`/case_new/:caseId`)
- Manage left sidebar auto-collapse
- Fix some TS issues coming from dependencies update


carere (approved):
Nice let's start from there :)

PR #757, opened 2025-03-25
feat: move github banner in LeftSidebar

ChibiBlasphem:
Moving github banner in the left sidebar to be less intrusive for the user
image


packages/app-builder/src/components/AstBuilder/edition/helpers.ts
```diff
@@ -114,6 +114,8 @@ export function groupByOperandType(
           )
         : [];
 
+      console.log(fieldOptions);
```
carere, line 117: Think about removing console logs ;)

carere (changes requested):
LGTM but need to remove console.logs

carere (approved):

PR #760, opened 2025-03-25
Api spec for data exploration in case manager

Pascal-Delange:
# Illustration of calling the data exploration API
Example call to client object listing endpoint

Call `POST https://api.checkmarble.com/client_data/transactions/list`
with body

```json
{
  "exploration_options": {
    "source_table_name": "accounts",
    "filter_field_name": "account_id",
    "filter_field_value": "00000000-0000-0000-0000-000000000000",
    "ordering_field_name": "transaction_at"
  },
  "cursor_id": "c978368b-4e24-44da-b8cb-73d1950132c0",
  "limit": 50
}
```

assuming we come from an object of type "accounts" with id `"00000000-0000-0000-0000-000000000000"` and we want to paginate after the transaction with object_id `"c978368b-4e24-44da-b8cb-73d1950132c0"` in the order of (transaction_at desc, object_id).

Assuming also a data model like

```json
{
  "tables": {
    "accounts": {
      "name": "accounts",
      "fields": {
        "object_id": {
          "data_type": "String",
          "name": "object_id",
          "nullable": false
        }
        // ...
      },
      "links_to_single": {},
      "exploration_options": [
        {
          // this one corresponds to an index created (pending creation) on transactions(account_id, transaction_at)
          // implicitly related to a "link to single" from transactions table.
          "source_table_name": "accounts", // this table
          "source_table_id": "...",
          "source_field_name": "object_id",
          "source_field_id": "...",
          "target_table_name": "transactions",
          "target_table_id": "...",
          "filter_field_name": "account_id",
          "filter_field_id": "...",
          "ordering_field_name": "transaction_at",
          "ordering_field_id": "...",
          "status": "valid"
          // ^ computed from index status in db
        }
      ]
    },
    "transactions": {
      "name": "transactions",
      "fields": {
        "object_id": {
          "data_type": "String",
          "name": "object_id",
          "nullable": false
        },
        "account_id": {
          "data_type": "String",
          "name": "account_id",
          "nullable": false
        },
        "transaction_at": {
          "data_type": "Timestamp",
          "name": "transaction_at",
          "nullable": false
        }
      },
      "links_to_single": {
        "account": {
          "id": "...",
          "parent_table_name": "account",
          "parent_table_id": "...",
          "parent_field_name": "object_id",
          "parent_field_id": "...",
          "child_table_name": "transactions",
          "child_table_id": "...",
          "child_field_name": "account_id",
          "child_field_id": "..."
        },
        "beneficiary": {
          "id": "...",
          "parent_table_name": "beneficiaries",
          "parent_table_id": "...",
          "parent_field_name": "object_id",
          "parent_field_id": "...",
          "child_table_name": "transactions",
          "child_table_id": "...",
          "child_field_name": "beneficiary_id",
          "child_field_id": "..."
        }
      },
      "exploration_options": []
    },
    "beneficiaries": {
      // ...
    }
  }
}
```

The call would return something like

```json
{
  "data": [
    {
      "object_id": "c978368b-4e24-44da-b8cb-73d1950132c0",
      "account_id": "00000000-0000-0000-0000-000000000000",
      "amount": 42,
      "transaction_at": "2025-03-24T00:000:00Z",
      "updated_at": "2025-03-24T00:000:00Z",
      "beneficiary": {
        // nested object from beneficiaries entry related to the transaction
        "object_id": "c978368b-4e24-44da-b8cb-73d1950132c0",
        "BIC": "SOGEFRPPXXX",
        "IBAN": "FR76...",
        "updated_at": "2025-03-24T00:000:00Z"
      }
    }
    // ... more transaction objects
  ],
  "pagination": {
    "next_cursor_id": "c978368b-4e24-44da-b8cb-73d1950132c0",
    "has_next_page": true
  }
}
```

# Some schemas to illustrate the data model exploration options
### layered data model

layered_data_model_exploration

### flat data model
flat_data_model_exploration

# Related PRs
- Implementation of data model DTO changes (exploration options) started in https://github.com/checkmarble/marble-backend/pull/934

# Status
- scaffolding of the main endpoints needed to explore end user data is ready
- some detailed points still need to be ironed out (e.g. how "secondary filters" work - see yaml file)
- implementation started in https://github.com/checkmarble/marble-backend/pull/934 for the data model related endpoints - I'll start work on the actual data reading endpoint (from case point of view) in a separate PR
- I'm in favor of a first review of work so far, even though I still keep it as draft

# Update status 10/04/2025
Implemented in 
https://github.com/checkmarble/marble-backend/pull/946
https://github.com/checkmarble/marble-backend/pull/942
https://github.com/checkmarble/marble-backend/pull/934

### DIFF
--- a/packages/marble-api/openapis/marblecore-api/cases.yml
+++ b/packages/marble-api/openapis/marblecore-api/cases.yml
@@ -438,6 +438,43 @@
         $ref: 'components.yml#/responses/403'
       '404':
         $ref: 'components.yml#/responses/404'
+/cases/{caseId}/pivot_objects:
+  get:
+    tags:
+      - Cases
+    summary: ->
+      Return the pivot objects present in a case, computed from the pivot values on decisions
+      in the case. Pivot objects are deduplicated and come with their actual content (if previously
+      ingested) if the pivot value is from an actual unique pivot "object" (not just a value on an entity).
+    operationId: getPivotObjectsForCase
+    security:
+      - bearerAuth: []
+    parameters:
+      - name: caseId
+        description: ID of the case to read from
+        in: path
+        required: true
+        schema:
+          type: string
+          format: uuid
+    responses:
+      '200':
+        description: Detailed values (depending on ingestion status) on
+        content:
+          application/json:
+            schema:
+              type: object
+              properties:
+                pivot_objects:
+                  type: array
+                  items:
+                    $ref: '#/components/schemas/PivotObjectDto'
+      '401':
+        $ref: 'components.yml#/responses/401'
+      '403':
+        $ref: 'components.yml#/responses/403'
+      '404':
+        $ref: 'components.yml#/responses/404'
 /cases/files/${caseFileId}/download_link:
   get:
     tags:
@@ -950,3 +987,56 @@ components:
             previous_value:
               type: string
               format: date-time
+    PivotObjectDto:
+      type: object
+      properties:
+        pivot_object_id:
+          type: string
+          description:
+            The "object_id" field of the pivot object. Can be null if the pivot type is "field" or if the pivot
+            does point to another unique field than "object_id", and the object has not been ingested yet.
+        pivot_value:
+          type: string
+          description: The actual pivot value, as on the decision. This value is used for grouping decisions.
+        pivot_id:
+          type: string
+          format: uuid
+        pivot_type:
+          type: string
+          enum: ['field', 'object']
+        pivot_object_name:
+          type: string
+          description: Name of the entity on which the pivot value is found.
+        pivot_field_name:
+          type: string
+          description: Name of the field used as a pivot value
+        is_ingested:
+          type: boolean
+          description: Whether the pivot object has been ingested or not (only for pivot type "object")
+        pivot_object_metadata:
+          type: object
+          description: Metadata of the pivot object, if it has been ingested (only for pivot type "object")
+          properties:
+            valid_from:
+              type: string
+              format: date-time
+          additionalProperties: true
+        pivot_object_data:
+          type: object
+          description: ->
+            Data of the pivot object, if it is a pivot object and it has been ingested (only for pivot type "object"),
+            otherwise {key:value} with the pivot field used. If it is an ingested object, may include nested objects
+            {link_name:{object}} where link_name is the name of a link pointing from the pivot object, and object is the
+            full data present on the object found following that link.
+          additionalProperties: true
+        number_of_decisions:
+          type: integer
+          description: Number of decisions that have this pivot value
+      required:
+        - pivot_value
+        - pivot_type
+        - is_ingested
+        - pivot_object_name
+        - pivot_field_name
+        - pivot_object_data
+        - number_of_decisions
--- a/packages/marble-api/openapis/marblecore-api/cases.yml
+++ b/packages/marble-api/openapis/marblecore-api/cases.yml
@@ -1041,10 +1041,14 @@ components:
           type: object
           required:
             - valid_from
+            - object_type
           properties:
             valid_from:
               type: string
               format: date-timeme
+            object_type:
+              type: string
+              description: object type in the data model that the data corresponds to
         data:
           type: object
           description: The actual data of the object, as described in the client data model.
--- a/packages/marble-api/openapis/marblecore-api/cases.yml
+++ b/packages/marble-api/openapis/marblecore-api/cases.yml
@@ -1013,22 +1013,8 @@ components:
         is_ingested:
           type: boolean
           description: Whether the pivot object has been ingested or not (only for pivot type "object")
-        pivot_object_metadata:
-          type: object
-          description: Metadata of the pivot object, if it has been ingested (only for pivot type "object")
-          properties:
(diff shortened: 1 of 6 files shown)


packages/marble-api/openapis/marblecore-api/ingestion.yml
```diff
@@ -92,6 +92,45 @@
         $ref: 'components.yml#/responses/401'
       '403':
         $ref: 'components.yml#/responses/403'
+/client_data/{tableName}/list:
```
Pascal-Delange, line 95: Logically it could join other routes under the "/ingestion" prefix, but I feel like that has over served its purpose already.

packages/marble-api/openapis/marblecore-api/ingestion.yml
```diff
@@ -134,3 +173,78 @@ components:
...
+                - type: string
+                - type: number
+            ordering_field_name:
+              type: string
+              description: The field on which we want to order the objects (in descending order)
+          required:
+            - source_table_name
+            - filter_field_name
+            - ordering_field_name
+        # secondary_filters:
```
Pascal-Delange, line 204: I'm thinking of adding a simple syntax like `field_name[ops]=value` in query params, where `ops` could be in `=,>,>=, ...".
Or use something "standard" like https://github.com/jirutka/rsql-parser. Do you have any opinion on the two ?

packages/marble-api/openapis/marblecore-api/ingestion.yml
```diff
@@ -134,3 +173,78 @@ components:
...
+          minimum: 1
+          maximum: 100
+          default: 30
+    ClientDataListResponseDto:
+      type: object
+      required:
+        - data
+        - pagination
+      properties:
+        data:
```
Pascal-Delange, line 228: what I'll add here is at least a preview from the object annotations, as implemented by @apognu in https://github.com/checkmarble/marble-backend/pull/931.

packages/marble-api/openapis/marblecore-api/cases.yml
```diff
@@ -882,3 +919,56 @@ components:
             previous_value:
               type: string
               format: date-time
+    PivotObjectDto:
```
Pascal-Delange, line 922: Here as well, I'll want to add at least a preview from the object annotations, as implemented by @apognu in https://github.com/checkmarble/marble-backend/pull/931.

packages/marble-api/openapis/marblecore-api/ingestion.yml
```diff
@@ -134,3 +173,78 @@ components:
             valid_from:
               type: string
               format: date-time
+    ClientDataListRequestBody:
```
Pascal-Delange, line 176: I figured that given the rather complex structure of the params that are passed, a json body is easier to deal with than a long list of untyped query params. Hence the POST endpoint for reading (listing) end user data.

apognu:
> // nested object from beneficiaries entry related to the transaction

What happens if the initial object already has a `beneficiary` field? Or more globally if there is a field named the same way as the table we need to nest?

Pascal-Delange:
> Are we allowed to be specify the order in the the exploration request? Wasn't the goal of defining the "ordering column" to create an index to use it for ordering? Letting this be changed at the request-level would defeat this, no?

Not sure if I understand correctly, but yes the "ordering column" should define _on which field_ we are allowed to order (on top of the base filter), of which in all generality there may be several (to be fair it should not happen too much 🤔 ). Though you could perhaps choose at request time if you want to filter in ASC/DESC order.

apognu:
Also, maybe it's the example, but we start from an account, and try to get all the transactions related to that account, but we still include the transaction's account in each of those objects, which arguably are going to all be the same as the object we started with?

Pascal-Delange:
> > // nested object from beneficiaries entry related to the transaction
> 
> What happens if the initial object already has a `beneficiary` field? Or more globally if there is a field named the same way as the table we need to nest?

Good question. A reasonable and perhaps good enough at first heuristic would be "show/return the nested object only if it does not clash with a field on the table".
A perhaps more complete solution is "return nested objects in a separate object", but it's not quite satisfactory unless we also say "only show one level of depth for related objects (otherwise it's the same question again).
A last and perhaps most general solution would be to wrap them in a `"__nested__"` object - in any case something that is unlikely to ever clash with the "main" fields.

A side question raised by this is, where do we best document the structure of this relatively free json object - clearly the openapi is not the right place. I'm thinking somewhere in the backend close to the endpoint definition, with perhaps a backlink here.

Pascal-Delange:
> Also, maybe it's the example, but we start from an account, and try to get all the transactions related to that account, but we still include the transaction's account in each of those objects, which arguably are going to all be the same as the object we started with?

Also good question.
I had thought about it and was considering to return only "related entities" that are neither the one we "come from", nor any parents thereof (if we display deeper related entities).
But that would be more of an implementation detail piloted by the backend.

Pascal-Delange:
@ChibiBlasphem @apognu FYI, I pushed a new commit reworking the data model for the pivot objects data we return, following the discussion in a not-thread above.

ChibiBlasphem (approved):
LGTM

PR #763, opened 2025-03-26
Fix Memory Leak on Firefox

carere:
For more info about the issue from radix-ui, here is an [interesting link](https://github.com/shadcn-ui/ui/issues/1668)

For more context:
- When we use a select inside a dialog on Firefox & Safari, we get a memory leak
- So we decided to use `MenuCommand` component from Marble in order to avoid this memory leak
- The issue about the select & the dialog is still pending on shadcn & radix side


siiick (approved):

Pascal-Delange (commented):
cool 👍  thanks for the investigation

PR #764, opened 2025-03-26
feat: improve ast builder operand menu fields section

ChibiBlasphem:
Fix some UI issues in Operand menu when field names were too long.

image


packages/app-builder/src/components/AstBuilder/edition/OperandMenu/DiscoveryList.tsx
```diff
@@ -88,21 +93,26 @@ export function DiscoveryList({ onSelect }: SmartMenuListProps) {
 
 type SubMenuFieldTriggerProps = {
   options: EnrichedMenuOption[];
-  path: string;
+  label: string;
+  depth: number;
 };
 function SubMenuFieldTrigger(props: SubMenuFieldTriggerProps) {
   const { t } = useTranslation('scenarios');
+  const padding = 24 + props.depth * 16; // p-6 + (p-4 * depth)
```
ChibiBlasphem, line 101: @carere @siiick Not sure what to do here, tailwind won't take dynamic classnames and I don't want to enumerate a limited number of depth -> classnames

carere: In Tailwind V4, it's possible to use string template but unfortunatly not in V3, so it is the right solution for me

carere (approved):

Pascal-Delange (commented):
nice ! 
To put my comment from DM in the open, would be great if we could just replace the From (except first level one) by something like "down then right arrow", or something smarter along the same idea 🤔

PR #769, opened 2025-03-31
feat(custom-lists): Display values count on lists page

siiick:
Display the values count on the custom lists page.

image


Pascal-Delange (approved):
LGTM

PR #775, opened 2025-04-04
Add new statuses for cases (keeping backward compatibility with old o…

apognu:
cc @carere


carere:
> cc @carere

I'll cherry pick this commit into my branch, thx for the updates, let me know when new content arrive, I'll add it to my branch, or you could add it directly to my branch `feat/new-decision-list` 😄

PR #777, opened 2025-04-06
fix: accept both UUID an short UUID in routes params

siiick:
- Updated the custom list edit page to accept both UUID and shortUUID in the URL parameter.
- Refactored the UUID-to-shortUUID conversion function to improve code clarity and maintainability.
- Updated the custom list breadcrumb to use shortUUID for consistency in links.

### DIFF
--- a/packages/app-builder/src/routes/_builder+/lists+/$listId.tsx
+++ b/packages/app-builder/src/routes/_builder+/lists+/$listId.tsx
@@ -22,10 +22,8 @@ import { downloadFile } from '@app-builder/utils/download-file';
 import useAsync from '@app-builder/utils/hooks/use-async';
 import { handleParseParamError } from '@app-builder/utils/http/handle-errors';
 import { REQUEST_TIMEOUT } from '@app-builder/utils/http/http-status-codes';
-import { parseParamsSafe } from '@app-builder/utils/input-validation';
+import { parseIdParamSafe } from '@app-builder/utils/input-validation';
 import { getRoute } from '@app-builder/utils/routes';
-import { shortUUIDSchema } from '@app-builder/utils/schema/shortUUIDSchema';
-import { fromUUIDtoSUUID } from '@app-builder/utils/short-uuid';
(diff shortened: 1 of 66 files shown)


packages/app-builder/src/routes/_builder+/lists+/$listId.tsx
```diff
@@ -49,18 +51,26 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
...
+  const parsedParams = await parseParamsSafe(
+    params,
+    z.object({
+      listId: z.union([shortUUIDSchema, z.string().uuid()]),
+    }),
+  );
+  if (!parsedParams.success) {
+    return handleParseParamError(request, parsedParams.error);
+  }
+  const customList = await customListsRepository.getCustomList(parsedParams.data.listId);
```
Pascal-Delange, line 64: question: does this imply that the repository is making the mapping of short to long uuid to call the backend ?

Pascal-Delange (commented):
thanks ! 
Could you please do the case page as well while you're at it ?

packages/app-builder/src/routes/_builder+/lists+/$listId.tsx
siiick: No, it doesn’t.
The shortUUIDSchema uses the sUUID-to-UUID transform utility function to validate and transform the parameter directly. If this schema fails, we fall back to a regular UUID string validator.
In the end, it always resolves to a proper UUID. The repository only handles actual list IDs.
```
export const shortUUIDSchema = z.string().transform((value, ctx) => {
  try {
    return toUUID(value);
```

carere (approved):

packages/app-builder/src/routes/_builder+/lists+/$listId.tsx
Pascal-Delange: ok got it, thanks

Pascal-Delange (approved):
👍  thanks !

PR #780, opened 2025-04-11
fix: fix incorrect form checking on sanction check whitelisting

ChibiBlasphem:
Checkboxes on firefox return "on"/"off" values instead of a boolean.


Pascal-Delange (approved):
any chance we have this in other places too ?

ChibiBlasphem:
> any chance we have this in other places too ?

Not that I know of.

PR #786, opened 2025-04-16
Add OpenAPI specs for new case events, get next case ID and get pagin…

apognu:
This adds the OpenAPI specs for:

* The new case events for assigning and SARs
* The "get next case ID" endpoint
* The "get paginated case decisions".

@ChibiBlasphem I saw that in #767, you already added the spec for the new "get cases with pivot" endpoint, so I didn't add it here.


apognu:
@ChibiBlasphem @carere Your turn to play. Tests will fail because of the new events not being implemented.

PR #793, opened 2025-04-24
feat: defer freshness call to client-side to avoid possible timeout breaks

ChibiBlasphem:
As said in the title, the call to data freshness was blocking for the request to load on the frontend, to avoid any issue with timeouts due to how remix works with defer the request is now done from the client side part.


carere (approved):
LGTM

PR #794, opened 2025-04-24
feat(aggregation filters): disabling fuzzy match filter temporarily

siiick:
Temporary workaround: This PR removes the fuzzy match filter from the selector UI to avoid issues caused by missing or broken database functions. It will be reintroduced once the DB-side implementation is fixed.


carere (approved):

siiick:
No longer needed, as we’re now handling PostgreSQL extensions properly.

Issue #803, opened 2025-04-29
"Related Table" incorrectly filled


Denny-g6labs:
Data model: Wrong binding

![Image](https://github.com/user-attachments/assets/63d4aa69-c44d-4871-8c82-f6aa9c87c2c7)

Seems like "Related Table" displays the wrong value. I expected "test_table".
I'm not sure if it's a bug in the frontend or in the backend.

PR #818, opened 2025-04-30
fix: replace field name by table name in link section of data model #803

ChibiBlasphem:
- parent field name was referenced instead of the related table


Pascal-Delange (approved):
👍

PR #819, opened 2025-04-30
feat(cases-manager): Add a all-my-cases filtered inbox view

siiick:
Appends a link to the case inboxes list to display all cases assigned to the user.

To avoid duplication and preserve the existing layout and logic of the inbox view, this new list is rendered as a standard inbox with an 'assigned-to-me' identifier. This allows us to fully reuse the existing logic. Only filters and navigation links are overridden as needed.

Feel free to challenge or suggest improvements to this approach.


ChibiBlasphem (approved):
LGTM

PR #828, opened 2025-05-05
fix: should shrink drawer when closing last tab on data exploration

ChibiBlasphem:
When closing last tab of data explorer we should shrink the drawer back to the pivot panel


Pascal-Delange (commented):
maybe I'm misunderstanding, but it looks like you pass a onLastTabClose prop and in the end you don't use it ?

Pascal-Delange (approved):

PR #832, opened 2025-05-05
fix(pagination): add missing cleanup function in useEffect

siiick:
For some reason, the pagination state was not updating when navigating between inboxes on the /inbox pages.
Adding a cleanup function to the useEffect() in useCursorPaginatedFetcher—which resets the pagination state—resolved the issue.


carere (approved):
LGTM 👍🏾

PR #835, opened 2025-05-05
fix: add empty state for pivot panel & discriminate pivots

ChibiBlasphem:
- Discriminate pivot objet on more than pivot values
- Add message to configure pivot when empty


carere (approved):
LGTM 👍🏾

siiick (approved):

PR #838, opened 2025-05-06
chore(eslint): Add i18next plugin

siiick:
Add [eslint-plugin-i18next](https://github.com/edvardchen/eslint-plugin-i18next) to `@marble/eslint-config` package.

See [disallow literal string (no-literal-string)](https://github.com/edvardchen/eslint-plugin-i18next/blob/main/docs/rules/no-literal-string.md#disallow-literal-string-no-literal-string) rule doc


siiick:
I will rework how we import the shared config to allow the @marble/ui-design-system package to ignore this new rule.

siiick:
we are moving to biome. Closing this one

PR #839, opened 2025-05-06
Remarks from QA

carere:
- [x] Allow Selecting tags while having the panel open
- [x] Remove left grey bar on Case Events when there is no filtered events
- [x] Fix weird gap when several pivot value on snooze panel
- [x] Fix double `dash` on score modifier 
- [x] Fix width of case alerts table
- [x] Fix decision status display on alerts table
- [x] Try new layout for the rule hit & trigger object columns
- [x] Add better colors for decision status & scenario name
- [x] Add limitations on number of rule & trigger object's fields to 3
- [x] Enable adding a file without a comment


packages/app-builder/src/components/Cases/RequiredActions.tsx
```diff
@@ -37,7 +37,7 @@ export const RequiredActions = ({
               })}
             >
               <Button variant="secondary" size="xs">
-                <span>Review pending sanction checks</span>
+                <span>Review screening hits</span>
```
Pascal-Delange, line 40: translation please

packages/app-builder/src/routes/ressources+/cases+/add-comment.tsx
```diff
@@ -23,7 +23,7 @@ const MAX_FILE_SIZE = MAX_FILE_SIZE_MB * 1024 * 1024;
 
 const schema = z.object({
   caseId: z.string().nonempty(),
-  comment: z.string().nonempty(),
+  comment: z.string(),
```
Pascal-Delange, line 26: that probably mostly works, but would still be great to not send the form if everything is empty - as it is the backend will gracefully ignore an empty input, but I wouldn't rely on this

packages/app-builder/src/components/Cases/CaseAlerts.tsx
```diff
@@ -78,34 +81,68 @@ export const CaseAlerts = ({
...
+                    map(([key, value]) => (
                       <span
                         key={key}
-                        className="border-grey-90 flex gap-1 rounded-sm border px-1.5 py-0.5 text-xs"
+                        className="border-grey-90 flex w-fit gap-1 truncate rounded-sm border px-1.5 py-0.5 text-xs"
                       >
-                        <span>{key}:</span>
-                        <FormatData data={parseUnknownData(value)} language={language} />
+                        {key === 'decisions-remains' ? (
+                          <span>+{value as number} more</span>
```
Pascal-Delange, line 101: Translation please.
Seriously please, we just talked about this.

packages/app-builder/src/components/Cases/CaseAlerts.tsx
```diff
@@ -78,34 +81,68 @@ export const CaseAlerts = ({
...
                 </div>
-                <div className="border-grey-90 flex h-0 min-h-full flex-col items-start gap-1 truncate border-r p-2">
-                  {Dict.entries(decision.triggerObject).map(([key, value]) => {
-                    if (['account_id', 'object_id', 'company_id'].includes(key)) return null;
-
-                    return (
+                <div className="border-grey-90 flex h-full flex-col items-start gap-1 overflow-hidden border-r p-2">
+                  {pipe(
+                    Dict.entries(decision.triggerObject),
+                    filter(([key]) => !['account_id', 'object_id', 'company_id'].includes(key)),
```
Pascal-Delange, line 87: The "N more" is pretty cool on the rule hits box, but I don't think it's useful in the trigger object display box (we never really care about the number of fields in there).
Just show a forth value instead.

packages/app-builder/src/components/Cases/CaseAlerts.tsx
```diff
@@ -78,34 +81,68 @@ export const CaseAlerts = ({
...
-                  {decision.ruleExecutions
-                    .filter((r) => r.outcome === 'hit')
-                    .map((r) => (
+                  {pipe(
+                    decision.ruleExecutions,
+                    filter((r) => r.outcome === 'hit'),
+                    (arr) => {
+                      const items = take(arr, MAX_ITEMS_DISPLAYED - 1);
+                      if (arr.length > MAX_ITEMS_DISPLAYED - 1) {
+                        // We add a fake rule execution to display the number of remaining executions
```
Pascal-Delange, line 119: I like the idea here - however, the implementation is not correct. It displays "+-1" when there are exactly 4 rules that hit (where it should just display the 4), and then going forward it displays "+0" where it should display "+2" (if 5 rules), and so on.

Pascal-Delange (changes requested):

packages/app-builder/src/routes/ressources+/cases+/add-comment.tsx
carere: It's already the case:
```ts
    // in the action fetcher
    if (data.comment !== '') {
      promises.push(cases.addComment({ caseId: data.caseId, body: { comment: data.comment } }));
    }
```
And I initialize the form with an empty string:
```ts
defaultValues: { caseId, comment: '', files: [] }
```

packages/app-builder/src/components/Cases/CaseAlerts.tsx
carere: You said you added translation, I did not want to overlap on your work. I will added then...

packages/app-builder/src/components/Cases/CaseAlerts.tsx
carere: Ok 👍🏾

packages/app-builder/src/components/Cases/CaseAlerts.tsx
Pascal-Delange: well I did, but this one is a completely new one introduced in this PR so risk of overlap is zero 😬

packages/app-builder/src/routes/ressources+/cases+/add-comment.tsx
Pascal-Delange: Ok I partially correct my comment: the backend is not called, but the remix action is called if I click the send button without a file and without a comment. I don't think this is necessary

carere: We use the action to send the file though

packages/app-builder/src/locales/fr/cases.json
```diff
@@ -189,5 +189,8 @@
   "decisions.rule.rule_group": "Groupe de règles",
   "decisions.outcome": "Résultat",
   "decisions.rules": "Règles",
-  "next_unassigned_case": "Aller à la prochaine investigation non-affectée"
+  "next_unassigned_case": "Aller à la prochaine investigation non-affectée",
+  "required_actions.decide_final_status": "Choisir le statut final",
+  "required_actions.no_more_pending_sanction_checks": "Check de sanction terminé",
+  "required_actions.review_screening_hits": "Check de sanctions ({{count}})"
```
Pascal-Delange, line 195:
```suggestion
  "required_actions.no_more_pending_sanction_checks": "Filtrage de sanction terminé",
  "required_actions.review_screening_hits": "Filtrage de sanctions ({{count}})"
```

Pascal-Delange (approved):
thanks ! 
I pushed one small trad fix, be careful not to lose it if you rebase & force push

packages/app-builder/src/locales/fr/cases.json
```diff
@@ -195,5 +195,8 @@
   "escalate-button.forbidden.hint.admin": "Vous devez configurer vers quelle boîte de réception les investigations de cette boîte sont escaladées.",
   "escalate-case.modal.callout": "En procédant à l’escalade de ce cas, il ne vous sera plus affecté et vous n’y aurez plus accès si vous n'êtes pas membre de la boîte de réception cible.<br/>Vous serez redirigé vers la boîte de réception actuelle.",
   "escalate-button.label": "Escalader",
-  "escalate-case.modal.submit-button.label": "Escalader"
+  "escalate-case.modal.submit-button.label": "Escalader",
+  "required_actions.decide_final_status": "Choisir le statut final",
+  "required_actions.no_more_pending_sanction_checks": "Check de sanction terminé",
+  "required_actions.review_screening_hits": "Check de sanctions ({{count}})"
```
Pascal-Delange, line 201:
```suggestion
  "required_actions.no_more_pending_sanction_checks": "Filtrage de sanction terminé",
  "required_actions.review_screening_hits": "Filtrage de sanctions ({{count}})"
```

PR #845, opened 2025-05-06
Fail early on invalid environment, default value for SESSION_MAX_AGE

apognu:
**Warning:** This PR is a breaking change related to the deployment!

When in production mode, it checks the environment for any missing or invalid variables and refuses to start if it finds an issue. It is quite contrary to what currently happens where the server would always start properly, and error out during the requests.

This is one approach. There could be better ones if we find crashing out a bit excessive, for instance, we could fail the healthcheck (the server would still start in probeless environments).

@ChibiBlasphem Please roast this to do it TheRightWay ™️ .


apognu:
For information, this only applies to production builds because Vite does not seem to execute the check on start (it probably waits for a first request).

Pascal-Delange (approved):
👍

PR #848, opened 2025-05-06
fix(case manager): use infinite loading in data explorer table pagination

ChibiBlasphem:


packages/app-builder/src/queries/client-object-list.ts
```diff
@@ -16,24 +16,26 @@ type UseClientObjectListQueryParams = {
...
-      const response = await fetch(endpoint(tableName, qs), {
+    queryFn: async ({ queryKey: [_r, __, tableName, qs], pageParam }) => {
+      const qsWithPageParam = pageParam ? `${qs}&offsetId=${pageParam}` : qs;
+      const response = await fetch(endpoint(tableName, qsWithPageParam), {
         method: 'GET',
       });
+
       return response.json() as Promise<DataTableObjectListResource>;
     },
+    initialPageParam: null as string | number | null,
```
carere, line 35: Is is necessary here ?

carere (approved):
LGTM, just have a little question 😃

PR #854, opened 2025-05-07
fix(case-manager): Prevent submit flooding

siiick:
- Make the event note mandatory.
- Validate the form on mount and on change.

### DIFF
--- a/packages/app-builder/src/routes/ressources+/cases+/add-comment.tsx
+++ b/packages/app-builder/src/routes/ressources+/cases+/add-comment.tsx
@@ -27,7 +27,7 @@ const MAX_FILE_SIZE = MAX_FILE_SIZE_MB * 1024 * 1024;
 
 const schema = z
   .object({
-    caseId: z.string().nonempty(),
+    caseId: z.string().uuid().nonempty(),
     comment: z.string(),
     files: z.array(z.instanceof(File)),
   })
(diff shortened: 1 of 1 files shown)


siiick:
My initial approach wasn’t 100% efficient.

@Pascal-Delange , could you please confirm whether we want to allow submitting files without an attached comment?

I originally tried to implement a form validation rule using .refine() to ensure either a comment or at least one file is provided:
```
const schema = z
  .object({
    caseId: z.string().nonempty(),
    comment: z.string(),
    files: z.array(z.instanceof(File)),
  })
  .refine(
    (data) => data.comment.trim() !== '' || data.files.length > 0,
    {
      message: 'You must provide either a comment or at least one file.'
    }
  );
```
However, I couldn’t get the validation to trigger properly after selecting a file — even when triggering it manually.

After checking Figma again, I didn’t find any examples of submitting a file without a note, so maybe simply requiring a non-empty note would be enough.

Pascal-Delange:
@siiick yes, we are fine with files uploaded without a comment

siiick:
> @siiick yes, we are fine with files uploaded without a comment

Ok so I need to try again with`refine()`. Maybe @carere or @ChibiBlasphem has a thought on it ?

packages/app-builder/src/routes/ressources+/cases+/add-comment.tsx
```diff
@@ -209,14 +219,23 @@ export function AddComment({ caseId }: { caseId: string }) {
...
       </div>
-      <Button
-        type="submit"
-        variant="primary"
-        size="medium"
-        aria-label={t('cases:case_detail.add_a_comment.post')}
-      >
-        <Icon icon="send" className="size-5" />
-      </Button>
+      <form.Subscribe selector={(state) => [state.canSubmit, state.isSubmitSuccessful]}>
```
carere, line 222: Why do you use `form.Subscribe` ?

packages/app-builder/src/routes/ressources+/cases+/add-comment.tsx
```diff
@@ -25,11 +25,13 @@ import { z } from 'zod';
...
 const MAX_FILE_SIZE = MAX_FILE_SIZE_MB * 1024 * 1024;
 
-const schema = z.object({
-  caseId: z.string().nonempty(),
-  comment: z.string(),
-  files: z.array(z.instanceof(File)),
-});
+const schema = z
+  .object({
+    caseId: z.string().nonempty(),
```
ChibiBlasphem, line 30: Didn't see that before (I know it's not from your changes), but I think it would be better with a `z.string().uuid()`

ChibiBlasphem (approved):
LGTM (the change asked is a nice to have)

packages/app-builder/src/routes/ressources+/cases+/add-comment.tsx
ChibiBlasphem: Tanstack form doesn't re-render the DOM by default, if you want to render something dependending on the state of the form (like if it's submitting or not) you need to use form.Subscribe (cf documentation: https://tanstack.com/form/latest/docs/framework/react/guides/reactivity#formsubscribe)

PR #859, opened 2025-05-09
Remove inbox settings from inbox members.

apognu:
Hopefully, this is the last fix on inbox management. We need to filter out inboxes where the user is only a member, since there is no real value in showing them.


ChibiBlasphem (approved):
LGTM

PR #860, opened 2025-05-12
chore: Move to biome

siiick:


biome.json
```diff
@@ -0,0 +1,85 @@
...
+    "formatWithErrors": false,
+    "attributePosition": "auto",
+    "indentStyle": "space",
+    "indentWidth": 2,
+    "lineWidth": 100,
+    "lineEnding": "lf"
+  },
+  "linter": {
+    "enabled": true,
+    "rules": {
```
carere, line 32: You forgot the rules  for TailwindCSS:

```json
    "rules": {
      "recommended": true,
      "nursery": {
        "useSortedClasses": {
          "fix": "safe",
          "level": "error",
          "options": {
            "attributes": ["class", "classList"],
            "functions": ["clsx", "cva", "twMerge"]
          }
        }
      }
    }
```

carere: You can also add a domain to add all rules related to it https://next.biomejs.dev/linter/domains/

PR #865, opened 2025-05-12
fix(case-manager): Reset component state

siiick:
Pass a key attribute to the PivotsPanelContent component to force its state to reset when navigating from one case to another.
see https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes


Pascal-Delange (approved):
🥲  thank you

PR #869, opened 2025-05-13
fix(inbox list): remove create inbox button when there's at least 1 inbox and show settings button MAR-946

ChibiBlasphem:
Marble · 12 53pm · 05-13


Pascal-Delange (commented):
👍 
(we agree that the title of the PR is incorrect however, right ?)

Pascal-Delange (approved):

ChibiBlasphem:
> 👍 (we agree that the title of the PR is incorrect however, right ?)

Yes it is, i'm changing it

PR #871, opened 2025-05-13
feat(case-manager): allow closing cases without selecting an outcome

siiick:


packages/app-builder/src/routes/ressources+/cases+/close-case.tsx
```diff
@@ -18,15 +18,15 @@ import { Button, cn, Modal } from 'ui-design-system';
...
+
 const schema = z.object({
-  caseId: z.string(),
-  outcome: z.union(
-    finalOutcomes.map((o) => z.literal(o)) as [
-      z.ZodLiteral<FinalOutcome>,
-      z.ZodLiteral<FinalOutcome>,
-      ...z.ZodLiteral<FinalOutcome>[],
-    ],
-  ),
```
carere, line 29: Can you use `z.union().optional()` ? Or `z.optional(z.union(...))` ?

siiick: I'm not sure we need a `Zod.union()` here.
[Zod.enums()](https://zod.dev/?id=zod-enums) fits best our needs since we have `finalOutcomes` which is an array :
```
export const finalOutcomes: UnionToArray<FinalOutcome> = [
  'false_positive',
  'valuable_alert',
  'confirmed_risk',
];
```

Pascal-Delange (approved):

PR #872, opened 2025-05-13
fix(case manager): rework right panel/drawer css

ChibiBlasphem:
Improve scroll behaviour on case manager page (independent scrollong main and panel, sticky sentinels and visual cues)


carere (approved):
Nice improvement, happy to finally see this different scrollable areas on this page 😄

PR #879, opened 2025-05-15
Add Content Security Policy

apognu:


packages/app-builder/src/routes/_auth+/sign-in.tsx
```diff
@@ -22,42 +23,27 @@ export const handle = {
...
-    getSignupStatus(),
-  ]);
+  const [err, appConfig] = await tryit(() => appConfigRepository.getAppConfig())();
 
-  if (ssoResponse.status === 'rejected') {
-    console.error('Error fetching internal SSO-enabled API');
-  }
-  if (signInStatusResponse.status === 'rejected') {
-    console.error('Error fetching internal signup-status API');
+  if (err) {
```
siiick, line 34: Silly question: Should we consider tryit a standard?
I’d naturally lean toward using a regular try/catch block in this case, but I’m okay with adopting this method more if that’s the direction we’re going.

siiick (approved):
Looks gorgeous to me !

packages/app-builder/src/routes/_auth+/sign-in.tsx
apognu: Error values > Exceptions 🥷

Pascal-Delange: let the religion wars begin 🙊 
joke aside, I rather like go's error values, but I wonder too if it's idiomatic in js 🤔

Pascal-Delange (approved):
really cool 👍

packages/app-builder/src/routes/_auth+/sign-in.tsx
ChibiBlasphem: I wasn't really opinionated before needing to no repeat code uselessly.

PR #880, opened 2025-05-16
feat(inboxes): rework filters

siiick:
🛠 Rework of the Inboxes Page Filters UI

Changes:
 - Removed the first headline containing filter information and consolidated all filter-related UI into the second headline.
 - The second headline is now always visible.
 - Introduced a new SimpleFilter entity and component for filters that don’t require additional settings.
 - When no filters are applied, a “Filters” button is shown on the left side of the filter bar.
When filters are active, a “New filter” button appears after the list of applied filters instead.


Pascal-Delange (approved):
let's merge !

PR #887, opened 2025-05-20
New OpenAPI spec for suspicious activity reports.

apognu:


packages/app-builder/src/components/Cases/CaseFile.tsx
```diff
@@ -13,6 +13,7 @@ import { casesI18n } from './cases-i18n';
 
 type CaseFileProps = {
   file: { id: string; fileName: string };
+  endpoint?: string;
```
Pascal-Delange, line 16: what is this ?

packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx
```diff
@@ -66,53 +98,34 @@ export async function action({ request }: ActionFunctionArgs) {
...
+          getServerEnv('MARBLE_API_URL_SERVER'),
+        ),
+        {
+          method: data.reportId ? 'PATCH' : 'POST',
           body,
           headers: { Authorization: `Bearer ${token}` },
-        }),
+        },
       );
+    } else if (data.reportId) {
```
Pascal-Delange, line 122: isn't there a "if method = DELETE" missing ? Even if it's maybe implicitly already the case due to the chaining of conditions.

packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx
```diff
@@ -66,53 +98,34 @@ export async function action({ request }: ActionFunctionArgs) {
...
 
-      promises.push(
-        fetch(new URL(path, getServerEnv('MARBLE_API_URL_SERVER')), {
-          method: 'POST',
+      body.append('caseId', data.caseId);
+      body.append('status', data.status);
+      if (data.file) body.append('file', data.file);
+      if (data.reportId) body.append('reportId', data.reportId);
+
+      await fetch(
```
Pascal-Delange, line 109: maybe I'm missing something, but we should be able to use the regular api client on the SSR server side to do this? (instead of crafting a fetch request with headers etc).

packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx
```diff
@@ -168,6 +201,10 @@ export const EditCaseSuspicion = ({
       }
     | undefined;
 
+  useEffect(() => {
+    console.log('Reports', reports);
```
Pascal-Delange, line 205: 🔥

Pascal-Delange (commented):
not done reading packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx which is a bit involved - here are a few questions already, I'll finish tomorrow (as we as doing a full review of translations before merging)

packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx
carere: Nope, because the regular client add the `multipart/form-data` to the request, and it should not, due to this handling, the server will not accept the request. It's a known problem with fetch https://flaviocopes.com/fix-formdata-multipart-fetch/

packages/app-builder/src/components/Cases/CaseFile.tsx
carere: Something I forgot

packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx
carere: Nope, it's all labout status & reportId, the condition are correct :)
If the status is "none", it means we want to delete the SAR, so I check if the reportId is present (and it **should** always be), but I prefer to check that.

packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx
Pascal-Delange: got it 👍 
As discussed orally, let's move that logic to the repo

Pascal-Delange (commented):
some unstaged comments I had from yesterday

packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx
```diff
@@ -194,6 +231,8 @@ export const EditCaseSuspicion = ({
     }
   }, [lastData]);
 
+  const reportFile = useStore(form.store, (state) => state.values.file);
+
```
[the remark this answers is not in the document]
carere: My bad, did not see that we already had the traduction key for the accepted extensions :) I fixed it ;)

Pascal-Delange (approved):
let's merge

PR #891, opened 2025-05-20
Pascal/pagination numbers fire

Pascal-Delange:


Pascal-Delange:
Apparently this passed the "early slack design review", so pushing this for code review now.

packages/app-builder/src/routes/_builder+/decisions+/_index.tsx
```diff
@@ -270,42 +271,6 @@ function SearchById() {
   );
 }
 
-// Temporary disabled
```
apognu, line 273: Temporary since November, 2023. 😆

apognu:
We might want to also display the seconds (or change the formatting some way) _if_ the first and last elements are within the same minute.

It is quite likely that an organizaiton will have more than 25 (or even more than 50) decision made within the same minutes, which would mean several pages would get the same pagination formatting, which in my opinion WouldBeBad.

packages/app-builder/src/routes/_builder+/decisions+/_index.tsx
Pascal-Delange: no such thing as temporarily commented code :D

Pascal-Delange:
You're right, will add tonight (or whenever I finish the PR)

Pascal-Delange:
Capture d’écran 2025-05-23 à 17 21 44
Capture d’écran 2025-05-23 à 17 21 37
Ok, new version where we show more precise time if it's in the same minute

packages/app-builder/src/components/PaginationButtons.tsx
```diff
@@ -50,16 +95,28 @@ export function CursorPaginationButtons({
...
+  console.log('endFormatted', endFormatted);
   return (
     <div className="flex items-center justify-end gap-2">
-      <Trans
-        t={t}
-        i18nKey="common:items_displayed"
-        components={{ StartToEnd: <span style={{ fontWeight: 'bold' }} /> }}
-        values={{ start, end }}
-      />
+      {hideBoundaries ? null : endFormatted === null ? (
```
Pascal-Delange, line 105: I actually hate nested ternaries, but js somehow pushes me to do it

Pascal-Delange:
I'm getting this error however

siiick:
I pushed commit [0e5a3035](https://github.com/checkmarble/marble-frontend/pull/891/commits/0e5a3035a7759e99f7358f6b352c7ebd6cc6a7e6) with the following changes:
 - Refactored the getStartAndEndFormatted function into a FormattedDatesRange component that handles both formatting and rendering of date ranges.
 - Introduced distinct wordings for: Same second / Same day / Different days
This allows us to render ranges like:

> From 23 Apr 2025, 10:52 to 28 Apr 2025, 10:53
> On 28 Apr 2025 from 10:52 to 10:53
> On 28 Apr 2025 at 10:53

 - Simplified the root component and clarified separation of concerns.

Pascal-Delange:
The last commit is a refactoring proposition, read it separately.
There was a formatDateTime util function, that included default presets for datetime display.
In the end, half of the places where it was used were passing some of the default props as undefined to suppress them, also the default options (timeformat: "short") are incompatible to mix with numeric formats ("year": "2-digit"...), so I figured it's really more simple and explicit to just let whoever is using the function, choose what format is reasonable for them.

siiick (approved):
LGTM 👍

PR #897, opened 2025-05-23
Pascal/sar trads 2

Pascal-Delange:
I went over the translations again, I'm mostly happy with this (except, need to check capitalization again).

It was... hard to work with the code however, mostly because of how the pages mixes the initial component state and form state. 

I also pimped the modale a bit. However, I'm not super happy with how complex this inline modale has become (different behavior & translations depending on current and target status...).

### Screenshots

- initial state
Capture d’écran 2025-05-23 à 09 32 30

- marked to do
Capture d’écran 2025-05-23 à 09 32 34

- modale for direct "mark completed" - pre-file choice
Capture d’écran 2025-05-23 à 09 32 47

- modale for direct "mark completed" - post-file choice
Capture d’écran 2025-05-23 à 09 32 54

- completed state with file
Capture d’écran 2025-05-23 à 09 33 00

- completed state without file
Capture d’écran 2025-05-23 à 09 33 12

- upload file modale post-mark as completed
Capture d’écran 2025-05-23 à 09 33 19

### and in french
Capture d’écran 2025-05-23 à 09 40 23
Capture d’écran 2025-05-23 à 09 40 16
Capture d’écran 2025-05-23 à 09 40 07
Capture d’écran 2025-05-23 à 09 40 02


packages/app-builder/src/repositories/CaseRepository.ts
```diff
@@ -164,8 +173,16 @@ export function makeGetCaseRepository() {
...
     createSuspiciousActivityReport: async ({ caseId, body }) =>
       adaptSuspiciousActivityReport(await marbleCoreApiClient.sarCreate(caseId, body)),
-    updateSuspiciousActivityReport: async ({ caseId, reportId, body }) =>
-      adaptSuspiciousActivityReport(await marbleCoreApiClient.sarUpdate(caseId, reportId, body)),
+    updateSuspiciousActivityReport: async ({ caseId, reportId, body }) => {
+      console.log('updateSuspiciousActivityReport', {
+        caseId,
+        reportId,
+        body,
+      });
```
carere, line 181: 🔥

packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx
```diff
@@ -217,11 +219,12 @@ export const EditCaseSuspicion = ({
   useEffect(() => {
     if (lastData?.success) {
       setOpenReportModal(false);
+      form.setFieldValue('reportId', lastData.data?.id);
+      setIsCompleted(lastData.data?.status === 'completed');
     }
-  }, [lastData]);
+  }, [form, lastData]);
```
carere, line 225: Don't follow React warning here, do not put the form in the dependency array, because it may go in infinite loop, if the form is modified and it will 😅

packages/app-builder/src/repositories/CaseRepository.ts
```diff
@@ -44,6 +43,16 @@ export type CaseFilters = {
...
+// file MUST not be present as an undefined value, because it will send "undefined" as a file
+type SuspiciousActivityReportBody =
+  | {
+      status?: 'pending' | 'completed';
+    }
+  | {
+      status?: 'pending' | 'completed';
+      file: Blob;
+    };
+
```
carere, line 55: You do not need to do that, since I already handle this in the action function on the component.
I just don't add the file to the body if it is undefined, that how I fixed this issue 😄

carere (changes requested):
Little modifications are needed, but I'm generally fine with those modifications :)

packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx
carere: If you use VsCode, you can disable the eslint rule for this line, we do this in other places in the app 😉

packages/app-builder/src/repositories/CaseRepository.ts
Pascal-Delange: oopsies

packages/app-builder/src/routes/ressources+/cases+/edit-suspicion.tsx
Pascal-Delange: alright thanks

packages/app-builder/src/repositories/CaseRepository.ts
Pascal-Delange: I know, this line here is just to clarify to ts that I actually want the key with a file, or don't want the key at all, but I don't want it to be present with undefined value.
I think this is the correct type (better than the generated DTO one) that goes along with the implem you did in the actual action ?

carere (approved):
LGTM 😄

PR #898, opened 2025-05-23
Add time to list of previous case & Id to Decision Panel

carere:
Screenshot From 2025-05-23 13-06-15

I decided to add the decision ID to the panel into this PR

Screenshot From 2025-05-23 13-31-55


Pascal-Delange (approved):
great ! lgtm

PR #902, opened 2025-05-27
feat(data-model): Hide not linked rows

siiick:
Hide unlinked fields in table model view by default

Introduces a filter to display only fields with links in the table model node, improving focus on relevant data relationships. Adds a toggle to show or hide unlinked fields and displays a message when no linked fields are present. Enhances usability for users working with complex data models.

### DIFF
--- a/packages/app-builder/src/components/Data/TableModelNode.tsx
+++ b/packages/app-builder/src/components/Data/TableModelNode.tsx
@@ -16,16 +16,20 @@ import { useDataModelFeatureAccess } from '@app-builder/services/data/data-model
 import { getRoute } from '@app-builder/utils/routes';
 import { NavLink } from '@remix-run/react';
 import {
+  type ColumnFiltersState,
   createColumnHelper,
   flexRender,
   getCoreRowModel,
+  getFilteredRowModel,
   useReactTable,
 } from '@tanstack/react-table';
 import clsx from 'clsx';
 import * as React from 'react';
+import { useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { Handle, type NodeProps, Position } from 'reactflow';
 import * as R from 'remeda';
+import { Button } from 'ui-design-system';
 import { Icon } from 'ui-icons';
 
 import {
@@ -56,6 +60,7 @@ export interface TableModelNodeData {
     isEnum: boolean;
     tableId: string;
     unicityConstraint: UnicityConstraintType;
+    hasLink: boolean;
   }[];
 }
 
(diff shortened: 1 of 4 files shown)


packages/app-builder/src/components/Data/TableModelNode.tsx
```diff
@@ -64,15 +69,23 @@ export function adaptTableModelNodeData(
   dataModel: DataModel,
   pivots: Pivot[],
 ): TableModelNodeData {
+  const linksToThisTable = dataModel
+    .filter((table) => table.id !== tableModel.id)
```
Pascal-Delange, line 73: technically it's possible (for now at least) to have self-referencing links, maybe we can just ignore this line ?

packages/app-builder/src/components/Data/TableModelNode.tsx
```diff
@@ -245,7 +286,29 @@ export function TableModelNode({ data }: NodeProps<TableModelNodeData>) {
               </tr>
             );
           })}
+
+          {table.getRowModel().rows.length === 0 ? (
+            <tr>
+              <td colSpan={columns.length + 2} className="text-grey-80 p-4 text-center">
+                {t('data:table.no_linked_fields.label')}
```
Pascal-Delange, line 293: This edge case is a good point, I hadn't thought of it.
I'm wondering if we should perhaps display a minimum of 1-2 fields with an expandable "..." if there are no links at all ?
Give me your opinion first before starting to implement though, this is just an initial idea from having a look at the screenshot.

Pascal-Delange (commented):
let's try to replace the icon used ("arrow down") by something more like an "expand" button: I'm thinking of something like "arrow up above arrow down", if you see what I mean).

packages/app-builder/src/components/Data/TableModelNode.tsx
siiick: `linkToThisTable` was extracted from the return value of `adaptModelNodeData`, which is used in the unique fields edition rules.
If we ignore this line, it would imply that self-referencing links can not be edited anymore.
That said, maybe we currently have a bug — I’m not entirely sure what the business rule is regarding the editing of self-referenced unique fields.
Unless the variable name is misleading and I’m misunderstanding the intent?

packages/app-builder/src/components/Data/TableModelNode.tsx
siiick: I’m in favor of always displaying the two mandatory fields (`object_id` and `updated_at`).
Are you suggesting we show a [...] button instead of a [^] one when there are no linked fields?

carere (approved):

PR #903, opened 2025-05-27
Add filtering on sanction list on the screening configuration

apognu:


Pascal-Delange:
Thanks,
Round 2 of UI/usage review (I'd like to let someone else review the code).
It's pretty cool like this, with many 3 minor points:
- "select all" should ideally only select the options that are currently visible (based on filters that have been applied), IMO - if it's not too time consuming to do
- the full text search has the right feeling UI-wise but... it just doesn't seem to filter quite like it should (see screenshot below)

Capture d’écran 2025-06-02 à 22 40 53

carere:
@Pascal-Delange I fixed the regression, but for the empty list placeholder, it will be on the next iteration on this 😉

carere (approved):

PR #909, opened 2025-06-05
Multiple screening rules per scenario.

apognu:


packages/app-builder/src/components/Scenario/Sanction/FieldBlackListId.tsx
```diff
@@ -0,0 +1,78 @@
...
+}
+
+export const FieldBlackListId = ({
+  value,
+  onBlur,
+  onChange,
+  editor,
+  customLists,
+}: FieldBlackListIdProps) => {
+  const { t } = useTranslation();
```
siiick, line 26: issue: we miss the desired namespace

```suggestion
  const { t } = useTranslation(['scenarios']);
```

packages/app-builder/src/components/Scenario/Sanction/FieldSkipIfUnder.tsx
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export const FieldSkipIfUnder = ({
+  value,
+  onBlur,
+  onChange,
+  editor,
+  name,
+}: FieldSkipIfUnderProps) => {
+  const { t } = useTranslation();
```
siiick, line 20: issue: we miss the desired namespace
```suggestion
  const { t } = useTranslation(['scenarios']);
```

siiick (changes requested):

PR #929, opened 2025-06-16
feat(data explorer): make head row sticky

ChibiBlasphem:
Make the column names row of the table sticky
image

At the moment this is the best solution I found to do it, but I think the right panel needs an overall rework for it to be able to manage sticky elements without as much hardcoded values


Pascal-Delange (approved):
Let's merge this ! 
We can rediscuss improvements on the structure later

PR #939, opened 2025-06-20
chore: move to biome v2

siiick:
Migrate from ESLint/Prettier to Biome v2
	•	Drop the @marble/eslint-config package
	•	Replace Prettier with Biome’s JS API in code generation scripts
	•	Update CI to use Biome
	•	Use rules similar to our existing setup to minimize impact in this first iteration

See sub-PR [#940](https://github.com/checkmarble/marble-frontend/pull/940) for the full codebase impact.

PR #940, opened 2025-06-20
chore: biome v2 lint

siiick:
Lint / Format existing codebase
following https://github.com/checkmarble/marble-frontend/pull/939


.vscode/.user-settings.sample.json
carere, line 1: Can you put your config as user setting that way we don't add this to the general codebase (I don't use mise)

carere (changes requested):

.vscode/.user-settings.sample.json
siiick: ok I will do the change in the #939 PR

PR #941, opened 2025-06-20
[Multiple Screening] Closing Select upon selection

carere:


Pascal-Delange (commented):
Thanks, looks good
Just one thing, there was the same issue on the custom list selector for removing terms from the input

Pascal-Delange (approved):

PR #942, opened 2025-06-20
fix(build): use legacy pnpm deploy behavior

siiick:
> By default, the deploy command only works with workspaces that have the inject-workspace-packages setting set to true. If you want to use deploy without "injected dependencies", use the --legacy flag or set force-legacy-deploy to true.

https://pnpm.io/cli/deploy


siiick:
> This seems to be the simple fix, but what does the "injected dependencies" do, and should we use it ? What is the change compared to the previous version ?

I'm still not sure, the documentation does not elaborate much on it. Still looking for details

Pascal-Delange (approved):

PR #943, opened 2025-06-20
[Multiple Screening] Last remarks from QA

carere:
- [X] Refine Modal broken due to decision id insted of sanction check id
- [X] Name recognition is now under feature access and flagged as beta
- [X] Now specifying "All Selected" for the list of dataset if a section is fully selected
- [X] Renaming "Sanction Check" to "Screening"
- [x] Warning the user about selected at least some attributes depending on some entities kind


Pascal-Delange (approved):
Thanks ! Let's merge :-)

PR #955, opened 2025-06-24
(refactor) Validation `onChange` & `onBlur` only for field,  and `onSubmit` for form

carere:
This PR remove the globals validators from all the Form in the app (except for the `onSubmit`).
The removed validators are now assigned to each form's field.


Pascal-Delange:
I'm arriving after the merge but naively this looks pretty verbose - is there any more "shorthand" way of handling this ?

ChibiBlasphem:
Is there a reason why all the validators have been moving to field level? I can't understand why we would want to have so much boilerplate per field.
When I tested it, just removing the `onBlur` form validation was working (if you have an `onChange` validation the `onBlur` is pretty much useless)

PR #960, opened 2025-06-30
feat(case manager): display decision review comment on case event timeline MAR-1102

ChibiBlasphem:
image


packages/app-builder/src/locales/fr/cases.json
```diff
@@ -161,6 +161,7 @@
   "required_actions.decide_final_status": "Choisir le statut final",
   "required_actions.review_screening_hits": "Filtrage de sanctions ({{count}})",
   "case_detail.history.event_detail.decision_reviewed": "Décision examinée à <Statut /> par <Actor>{{actor}}</Actor>",
+  "case_detail.history.event_detail.decision_reviewed_with_comment": "Décision examinée avec un commentaire à <Statut /> par <Actor>{{actor}}</Actor>",
```
Pascal-Delange, line 164:
```suggestion
  "case_detail.history.event_detail.decision_reviewed_with_comment": "Décision examinée avec un commentaire à <Status /> par <Actor>{{actor}}</Actor>",
```

Pascal-Delange (changes requested):
- small typo
- even if we iterate on this quickly afterwards, let's still add this new event type to the two that are displayed by default, for consistency

packages/app-builder/src/locales/fr/cases.json
Pascal-Delange: no worries - still planning to do a big "ai agent" run of "find me any inconsistencies in translations"

packages/app-builder/src/components/Cases/CaseEvents.tsx
```diff
@@ -51,7 +51,7 @@ export function CaseEvents({
   const [olderEvents, setOlderEventsCount] = useState(0);
   const [newerEvents, setNewerEventsCount] = useState(0);
   const [filters, setFilters] = useState<CaseEventFiltersForm>({
-    types: ['comment_added', 'file_added'],
+    types: ['comment_added', 'file_added', 'decision_reviewed'],
```
Pascal-Delange, line 54: you're going to have it but there are two more independent places where we have logic on `'comment_added', 'file_added'` => check out `packages/app-builder/src/components/Cases/Events/Filters/index.tsx`
I tried to refactor it a month or so ago but I failed to make typescript happy - perhaps you have more luck, but in any case it's probably better done in the follow-up PR we discussede

packages/app-builder/src/locales/fr/cases.json
ChibiBlasphem: I used cursor agent to do this one but he took the line above to help but the line already had an error (which I fixed)

Pascal-Delange (approved):
great !

PR #962, opened 2025-06-30
feat(case manager): reducing case events to categories

ChibiBlasphem:
image


Pascal-Delange:
Here is my proposal @ChibiBlasphem , let me know.
If it's good for you, I'll approve and we can merge.
Capture d’écran 2025-06-30 à 13 50 58

PR #969, opened 2025-07-02
feat(decisions): Improve data fetching

siiick:
This PR introduces the following changes:
	•	refactor: Extracts data fetching logic into a dedicated service module for better separation of concerns and reusability.
	•	feat: Improves performance by fetching the datasets collection on demand instead of eagerly loading it.
	•	feat: Increases system resilience by skipping the sanctions check data enrichment step when the datasets list is unavailable (e.g., Yente service is down), avoiding unnecessary failures.

These changes aim to make the data handling more modular, efficient, and fault-tolerant.

### DIFF
--- a/packages/app-builder/src/routes/_builder+/decisions+/$decisionId.tsx
+++ b/packages/app-builder/src/routes/_builder+/decisions+/$decisionId.tsx
@@ -22,6 +22,7 @@ import { isNotFoundHttpError, Pivot } from '@app-builder/models';
 import { DecisionDetails } from '@app-builder/models/decision';
 import { SanctionCheck } from '@app-builder/models/sanction-check';
 import { ScenarioIterationRule } from '@app-builder/models/scenario-iteration-rule';
+import { createDecisionDetailsService } from '@app-builder/services/decisions/details';
 import { initServerServices } from '@app-builder/services/init.server';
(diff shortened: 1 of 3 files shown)


siiick:
> Do we really need for a function factory here (createDecisionDetailsService)? Could it not be just pure functions?
The idea was to use dependency injection pattern to easily test the service. The factory is not a strong will though.
Have you something else in mind like higher-order function ?

siiick:
> Why is enrichSanctionCheck is in there if it's acting on a sanctionCheck with the sanctionCheckRepository?
I knew the function name was misleading since we have a "enrich" feature on this page :D
I will find a better name.

siiick:
> I don't know if it's all the try/catch which are bugging me or the verbose use of Sentry.captureException but it makes the code look a little "too much" and make me feel it could be cleaner without all of those.

I wanted to add more context to the Sentry events we send. I think we often lack of context information in Sentry.
You are right, maybe i rather should intercept the error, add some context into it. throw it again and let it be catched at and higher level (I mean at the app level)

Pascal-Delange (commented):
Hey,
I'd love to chat about this one of those days (and to try to finish the PR while it's somewhat fresh in memory).
I agree that it goes in the right direction of moving logic out of loaders, but that it's a bit cumbersome to manipulate

PR #987, opened 2025-07-21
Add endpoints and attributes for case auto-assignment.

apognu:


packages/marble-api/openapis/marblecore-api/users.yml
```diff
@@ -0,0 +1,35 @@
...
+    security:
+      - bearerAuth: []
+    requestBody:
+      required: true
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/PersonalSettingsUnavailableDto"
+    responses:
+      "204":
```
apognu, line 16: Add potential 400/422 on wrong dates.

apognu: Yeah I was waiting that we settle on how this endpoint works (which, as I understand, is till not definitive before adding them).

packages/marble-api/openapis/marblecore-api/users.yml
```diff
@@ -0,0 +1,66 @@
+/settings/me/unavailable:
+  get:
+    tags:
+      - Personal settings
+    summary: Get the current unavailability
+    operationId: setUnavailable
```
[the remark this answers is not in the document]
siiick: fixed in #933

PR #1002, opened 2025-07-29
keep "+" while decoding email param

Pascal-Delange:
Keep the "+" in the email received in query param, which may not have been properly encoded if received from the firebase email template.


ChibiBlasphem (approved):
Seems good to me, but I don't think we will ever receive a literal "+" in url as it's encoded in "%20"

PR #1008, opened 2025-08-05
feat: improve inboxusers endpoints

siiick:
### Goal

Replace the fetcher with queries to prevent full page reloads on each toggle switch action, which could previously lead to data validation mismatches if the user interacted too quickly.

### Refactor
	•	Moved inboxUsers-related resource endpoints (which were returning components) into a dedicated inbox-users directory to simplify maintenance.
	•	Created new, cleaner API-only endpoints under a separate inboxusers directory.

### API Contract Update
	•	Updated updateInboxUsers parameters: all fields are now optional to allow partial updates (i.e., update a single parameter without needing to provide all others).
This change aligns with the actual behavior of the API.

### DIFF
--- a/packages/app-builder/src/models/inbox.ts
+++ b/packages/app-builder/src/models/inbox.ts
@@ -123,8 +123,8 @@ export function adaptInboxUserUpdateBody(data: InboxUserUpdateBody): {
   auto_assignable?: boolean;
 } {
   return {
-    ...(data.role !== undefined && { role: data.role }),
-    ...(data.autoAssignable !== undefined && { auto_assignable: data.autoAssignable }),
+    role: data.role,
+    auto_assignable: data.autoAssignable,
   };
 }
 
--- a/packages/app-builder/src/routes/_builder+/settings+/inboxes.$inboxId.tsx
+++ b/packages/app-builder/src/routes/_builder+/settings+/inboxes.$inboxId.tsx
@@ -233,10 +233,10 @@ export default function Inbox() {
               size: 150,
               cell: ({ getValue, row }) => {
                 const [value, setValue] = useState(getValue());
-                const { mutate: editAutoAssign } = useEditAutoAssignMutation();
+                const editAutoAssignMutation = useEditAutoAssignMutation();
                 const handleChange = (checked: boolean) => {
                   setValue(checked);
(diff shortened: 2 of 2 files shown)


packages/app-builder/src/models/inbox.ts
```diff
@@ -114,17 +114,17 @@ export function adaptInboxUserCreateBody({
...
-  role: string;
-  auto_assignable: boolean;
+  role?: string;
+  auto_assignable?: boolean;
 } {
   return {
-    role: data.role,
-    auto_assignable: data.autoAssignable,
+    ...(data.role !== undefined && { role: data.role }),
+    ...(data.autoAssignable !== undefined && { auto_assignable: data.autoAssignable }),
```
ChibiBlasphem, line 127: Pourquoi les ternaires ? Il y avait un problème avec TS ?

packages/app-builder/src/routes/_builder+/settings+/inboxes.$inboxId.tsx
```diff
@@ -231,24 +232,14 @@ export default function Inbox() {
               header: t('settings:inboxes.inbox_details.auto_assign_enabled.label'),
               size: 150,
               cell: ({ getValue, row }) => {
-                const fetcher = useFetcher<typeof UpdateInboxUser>();
                 const [value, setValue] = useState(getValue());
+                const { mutate: editAutoAssign } = useEditAutoAssignMutation();
```
ChibiBlasphem, line 236: I would prefer if we use `const editAutoAssignMutation = useEditAussignMutation()` rather than spreading the object.
It makes it much more clear when you have multiple queries/mutations and to have a consistency between the uses of queries/mutations I think it's better to do it even if we have only one.

What do you think?

packages/app-builder/src/models/inbox.ts
siiick: It’s not a specific TypeScript issue—it’s just the easiest way I know in JavaScript to create an object with conditional keys.

ChibiBlasphem: Why do you want conditional keys if the value of is already undefined ?
For example: `JSON.stringify({ foo: undefined }) == "{}"`

PR #1013, opened 2025-08-11
Chore/migrate zod v4

siiick:
Migration to Zod v4


ChibiBlasphem (approved):
LGTM. Didn't read it thoroughly as it's mainly a codemod being ran. Typescript is ok so it's ok for me.

PR #1020, opened 2025-08-18
fix: remove ariakit's FormLabel

siiick:
CreateField component was using Ariakit's FormLabel component instead of tanStack one.


ChibiBlasphem (approved):
lgtm

PR #1024, opened 2025-08-19
feat(data-models): improve form validation and error handling

siiick:
- Disable field validation on blur to prevent error messages from appearing when users close the modal.
- Improve feedback when attempting to create a field with a duplicate name.
- Translate validation and error messages.
- Fix incorrect regex used for field name shape validation.

### DIFF
--- a/packages/app-builder/src/components/Data/CreateField.tsx
+++ b/packages/app-builder/src/components/Data/CreateField.tsx
@@ -6,9 +6,15 @@ import {
   createFieldValueSchema,
   useCreateFieldMutation,
 } from '@app-builder/queries/data/create-field';
+import {
+  type CreateFieldValidationErrorCode,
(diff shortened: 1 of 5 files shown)


packages/app-builder/src/components/Data/CreateField.tsx
```diff
@@ -49,6 +49,15 @@ function CreateFieldContent({ tableId, closeModal }: { tableId: string; closeMod
...
 
+  // Helper function to translate error messages
+  const translateError = (key: string) => {
+    try {
+      return t(key as any);
+    } catch {
+      return key; // Fallback to the key if translation fails
+    }
+  };
+
```
ChibiBlasphem, line 60: If it's a helper it would be cool to put it in utils (somewhere)

PR #1025, opened 2025-08-20
feat: improve locale handling

siiick:
Get applied language from tranlsation hook instead of from the navigator.


siiick:
@Pascal-Delange what are you thoughts on this ?

PR #1026, opened 2025-08-21
fix(workflows): correct validation for evluate string/numbers AST constant

siiick:
Constant AST nodes have `null` names. Adapt the Zod schema validation accordingly.


ChibiBlasphem (approved):
lgtm

PR #1027, opened 2025-08-21
fix(case-manager): unsnooze button

siiick:
- Change the Unsnooze button to directly perform the unsnooze action instead of displaying the snooze case menu.
 - Prevent users from selecting a past date in the custom date selector when snoozing a case.
 - ~~Automatically unsnooze a case immediately after a comment is posted if the case was snoozed.~~

Refs: MAR-1201


leo-marble:
Would it make sense to handle the unsnooze action when adding a comment on the backend to simplify the frontend logic? It feels like this behavior might belong more in the business logic on the backend rather than the frontend. Not sure, just wanted to get your thoughts on this.

ChibiBlasphem (approved):
lgtm

PR #1033, opened 2025-08-26
refactor(resources/cases): migrate cases resources fetchers

ChibiBlasphem:


siiick (approved):
My brain almost melted but LGTM :D

PR #1052, opened 2025-08-28
fix(rule snooze): fix invalid time value

ChibiBlasphem:
- Remove condition on decision.id being the same it can't be possible between cases (e.g: rule is snoozed based on ruleId and the pivot, not the decision id itself)


siiick (approved):

Pascal-Delange:
Je me permets de close vu que ça a été repris dans ma PR

PR #1055, opened 2025-08-28
fix on rule snooze & improvement

Pascal-Delange:
Open question on whether it's ok to export the PivotObjectDetails like I did, or if it should be moved somewhere else

### DIFF
--- a/packages/app-builder/src/components/CaseManager/SnoozePanel/SnoozePanel.tsx
+++ b/packages/app-builder/src/components/CaseManager/SnoozePanel/SnoozePanel.tsx
@@ -72,7 +72,6 @@ export const SnoozePanel = ({
                 </TabsList>
                 {Dict.entries(rulesByPivot).map(([pivotValue, rules]) => {
                   const client = findDataFromPivotValue(pivotObjects ?? [], pivotValue);
-                  console.log('client', client);
                   const table = dataModelWithTableOptions.find(
                     (t) => t.name === client?.pivotObjectName,
                   );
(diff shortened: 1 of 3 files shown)


packages/app-builder/src/components/CaseManager/SnoozePanel/SnoozePanel.tsx
```diff
@@ -60,28 +59,20 @@ export const SnoozePanel = ({
               <Tabs className="flex w-full flex-col" defaultValue={Object.keys(rulesByPivot)[0]}>
                 <TabsList className="mb-6 w-fit">
                   {Object.keys(rulesByPivot).map((pivotValue) => {
-                    const client = findDataFromPivotValue(pivotObjects ?? [], pivotValue);
-
-                    const clientName =
```
Pascal-Delange, line 65: this was complete nonsense

packages/app-builder/src/components/CaseManager/SnoozePanel/SnoozePanel.tsx
```diff
@@ -123,7 +116,6 @@ export const SnoozePanel = ({
                           <span className="p-2">{t('cases:decisions.rule.name_and_score')}</span>
                           <span className="p-2">{t('cases:decisions.rule.description')}</span>
                           <span className="p-2">{t('cases:decisions.rule.rule_group')}</span>
-                          <span className="p-2">{t('cases:decisions.outcome')}</span>
```
Pascal-Delange, line 126: This was completely misleading. Well, not the only thing that doesn't work in the design but it's the worst one.

ChibiBlasphem (approved):
LGTM!

PR #1059, opened 2025-08-29
fix(ast builder): literal list being coerced as string

ChibiBlasphem:
- List coercion wasn't supporting "non-word" characters insider the strings, make it possible to contain anything being not (`" , ] [`)


Pascal-Delange (approved):
Looks good to me, though I'm not the regexp expert

PR #1061, opened 2025-08-29
fix(ui-design-system): cast sharedTailwindConfig to Config type in tailwind config

siiick:
To prevent type mismatch


ChibiBlasphem:
Closing this PR as the problem was somewhere else

PR #1067, opened 2025-09-02
fix: prevent Nudge hover card to hide while hover

siiick:
The nudge hovercard was hiding even when the user hovered over it.
Fixed by removing its rendering as a React Portal.

image

Ref: MAR-1220


ChibiBlasphem (approved):
lgtm

PR #1076, opened 2025-09-05
feat(ip-whitelisting): add global error and settings UI MAR-1215

ChibiBlasphem:
IP Whitelisting settings, only accessible to admins:
image

In dev mode you can "simulate" your real IP by using `CUSTOM_X_REAL_IP` variable:
```
CUSTOM_X_REAL_IP=2.2.2.2 pnpm --filter app-builder run dev
```


siiick (approved):
LGTM

PR #1077, opened 2025-09-05
fix nudge display on settings/webhooks

Pascal-Delange:
## Issues
- "nudge" button is invisible if the side bar is folded (which should increasingly be its default status)
- "webhooks" nudge button in setting was not handling all the cases where it should be displayed
- "upgrade" button link in the hover did not do anything on click, only on "ctrl+click"

## Attempts
- I tried to make the nudge a miniature when the side bar is folded. I managed to make something that feels about right on the initial render, but the icon shifts back over the analytics page logo on the client re-render. I wasn't able to fix this.
- on the webhooks setting page, the nudge logo flickers around when we refresh the page, if the webhooks setting page is selected

server render:
Capture d’écran 2025-09-05 à 11 36 00
client render:
Capture d’écran 2025-09-05 à 11 32 44

Some of the changes in this PR have been made with cursor, handle with care (especially the css)

### DIFF
--- a/packages/app-builder/src/components/Navigation.tsx
+++ b/packages/app-builder/src/components/Navigation.tsx
@@ -14,6 +14,7 @@ export interface SidebarLinkProps {
   labelTKey: ParseKeys<['navigation']>;
   to: string;
   children?: React.ReactNode;
+  className?: string;
 }
 
 export const sidebarLink = cva(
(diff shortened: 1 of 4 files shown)


Pascal-Delange:
Can't approve my own PR, but LGTM!

PR #1090, opened 2025-09-05
chore: stage to have production node_module to reduce image size

OrriMandarin:
This pull request refactors the `Dockerfile` to optimize dependency management and build stages for both development and production environments. The main changes introduce a clear separation between installing development and production dependencies, which improves build efficiency and reduces the runtime image size.

**Dependency management improvements:**

* Added a dedicated `deps-prod` stage to install only production dependencies, separating them from the dev dependencies used in earlier build stages.
* Updated the runtime stage to copy only the production dependencies from the new `deps-prod` stage, ensuring that unnecessary dev dependencies are excluded from the final image.

**Build process clarifications:**

* Clarified comments and stage names throughout the `Dockerfile` to indicate which stages use dev dependencies and which use production dependencies, making the build flow easier to understand. [[1]](diffhunk://#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L1-R1) [[2]](diffhunk://#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557L16-R20)

**Artifact handling:**

* Changed the build output copy step to only transfer the necessary build artifacts, further reducing the size and complexity of the runtime image.


siiick (approved):

OrriMandarin:
Reduce from 1.6GB to 800MB

Pascal-Delange:
> Reduce from 1.6GB to 800MB

wait what, since when is the image size 1.6Gb ???
It used to be 400-ish Mb

Pascal-Delange:
That's a pretty brutal increase, I'd be happy to know more about what triggered it

OrriMandarin:
Capture d’écran 2025-09-05 à 15 46 33

@Pascal-Delange I have this listed in my Docker for the image sizes. We’re above what we had before bun, but not by much.

But there might be some optimizations to make, especially not including the test package, and maybe the way Remix is embedded to launch the server.

PR #1098, opened 2025-09-10
chore(build): update sentry organisation

siiick:
Fix Sentry org at build


Pascal-Delange (approved):
please remove it instead, I think it's redundant because it's already passed in the token

PR #1102, opened 2025-09-11
feat: display AI settings only for dev or managed environments

siiick:
Add the current Node environment as AppConfig.
Use it to condition the AI settings page display (Not available in production for non SaaS environments)


siiick:
@Pascal-Delange @ChibiBlasphem let's use the existing `AppConfig.versions.appVersion` instead.

PR #1107, opened 2025-09-15
chore(dependabot): Enable beta ecosystems support

siiick:
Dependabot currently does not update the Bun lockfile in its bump PRs.
This PR enables enable-beta-ecosystems: true in the Dependabot configuration, which allows Bun lockfiles to be updated automatically.

Source: https://github.com/dependabot/dependabot-core/issues/6528#issuecomment-3001386217

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -1,4 +1,5 @@
 version: 2
+enable-beta-ecosystems: true
 updates:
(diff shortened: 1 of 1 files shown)


OrriMandarin:
Ok! We need to follow the development of this feature. 
I saw in comment, this setting will remove dependabot reports (changelog), very sad :(

PR #1126, opened 2025-09-19
feat(WIP): new analytics page

siiick:
This PR introduces the foundation of the new in-app analytics page.
The page is still a work in progress, but it is safe to merge to production since it remains hidden behind a feature flag.

Key changes:
	•	New analytics page (WIP) with initial structure.
	•	Filters & selectors:
	•	Scenario selector
	•	Date range filter with support for a second range

Performance & architecture:
	•	Parallelized API requests
	•	Shifted business logic into the backend BFF for cleaner front-end code

Refs : MAR-1238


packages/app-builder/src/components/Analytics/Decisions.tsx
```diff
@@ -0,0 +1,207 @@
...
+            },
+          }}
+          //   markers={currentDataGroup?.scenarioVersionsXMarkers}
+        />
+        <div className="relative bottom-2 right-2 flex gap-2">
+          <ButtonV2
+            variant="secondary"
+            mode="normal"
+            onClick={() => setGroupDate('daily')}
+            className={groupDate === 'daily' ? 'bg-purple-98 border-purple-65 text-purple-65' : ''}
```
ChibiBlasphem, line 181: It's ok for now but maybe we should see with the designer if we eventually need a "selected/highlighted" version of the button...

ChibiBlasphem (changes requested):

ChibiBlasphem (approved):

PR #1134, opened 2025-09-24
feat: set default inbox page size to 50 and make it customizable MAR-1245

ChibiBlasphem:
- Make default inbox pagination limit 50
- Add button to customize limit for 25, 50 and 100 items per page
image

### DIFF
--- a/packages/app-builder/src/components/PaginationButtons.tsx
+++ b/packages/app-builder/src/components/PaginationButtons.tsx
@@ -13,7 +13,7 @@ export const paginationSchema = z.object({
   offsetId: z.string().optional(),
   next: z.coerce.boolean().optional(),
   previous: z.coerce.boolean().optional(),
-  limit: z.number().optional(),
+  limit: z.coerce.number().optional(),
   order: z.enum(['ASC', 'DESC']).optional(),
   sorting: z.enum(['created_at']).optional(),
 });
@@ -28,6 +28,7 @@ type CursorPaginationsButtonsProps = PaginatedResponse<ItemWithId> & {
   hasPreviousPage: boolean;
   boundariesDisplay: 'ranks' | 'dates';
   pageNb: number;
+  itemsPerPage?: number;
 };
 
 function FormattedDatesRange({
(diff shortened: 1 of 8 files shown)


packages/app-builder/src/components/PaginationButtons.tsx
```diff
@@ -143,13 +144,21 @@ function FormattedDatesRange({
...
+  pageNumber: number;
+  currentPageItemCount: number;
+  itemsPerPage: number;
+}) {
   const { t } = useTranslation(['common']);
 
-  const start = pageNb * defaultPaginationSize + 1;
-  const end = nbInPage > 0 ? pageNb * defaultPaginationSize + nbInPage : 0;
+  const start = pageNumber * itemsPerPage + 1;
+  const end = currentPageItemCount > 0 ? pageNumber * itemsPerPage + currentPageItemCount : 0;
```
Pascal-Delange, line 159: Simple clarification needed: if I understand correctly, in the end you left in place my hack of "rank displayed starts at pagination_size * page number", which is not really valid if the page size changed during pagination.
It's kind of a bug, but given it should happen quite rarely and even then affect only non-critical data, I may be ok introducing it if you believe the alternative would take more than 1h to do, namely:
- keep track of the number of items seen as we paginate on top of the number of pages seen.
- or something entirely different using the tanstack query technique you talked to me about on Tuesday

and provided that, in any case, we document the bug in linear and as a comment in this function

packages/app-builder/src/routes/_builder+/cases+/inboxes.$inboxId.tsx
```diff
@@ -77,6 +82,7 @@ export async function loader({ request, params }: LoaderFunctionArgs) {
 
   const parsedQuery = await parseQuerySafe(request, casesFiltersSchema);
   const parsedPaginationQuery = await parseQuerySafe(request, paginationSchema);
+  console.log('parsedPaginationQuery', parsedPaginationQuery);
```
Pascal-Delange, line 85: 🔥

packages/app-builder/src/components/PaginationButtons.tsx
Pascal-Delange: ok scratch that, do I understand correctly that changing the number of entries per page resets the pagination state completely ?

Pascal-Delange (approved):

PR #1227, opened 2025-12-01
chore: remove use fetcher frontend MAR-1343

ChibiBlasphem:


ChibiBlasphem:
Done on migration #1414

PR #1315, opened 2026-02-06
feat(continuous_screening): Expose continuous task status

OrriMandarin:
Linked: https://github.com/checkmarble/marble-backend/pull/1513


Pascal-Delange (commented):
Looks generally alright ! 
Now, I didn't get to try this out with pagination because not enough data in my local instance.
What's missing IMO before possibly merging this:
- clearer translations (non-technical translations)
- wondering if a good place for this is in settings or in a kinda hidden section in monitoring (I'd still say, the latter)
- human intellegible callouts/explanation texts that explain what is what

PR #1423, opened 2026-04-16
Remove OpenAPI v0 button o/.

apognu:
The venerable v0 is going to be deprecated, and all related API will be removed from the backend.

Keeping this in results in a Lose Your Marble when the backend returns 404.

I once again request that you plz mrg.

## Release Notes

* **Chores**
  * Removed deprecated OpenAPI v0 download functionality across the application; only OpenAPI v1 specification is now available for download. Locale entries updated accordingly in all supported languages.


ChibiBlasphem (approved):
lgtm

PR #1425, opened 2026-04-16
dark mode for flow, default semantic type on old tables and field creation

william-schlegel:
MAR-1737: Default selected type for Float should be Number
MAR-1772: [Subjective] Dark mode colors are not great
MAR-1771: Semantic type should not be guessed if it was manually set

## Summary by CodeRabbit

* **New Features**
  * Data flow visualization now respects the app's current theme settings.

* **Bug Fixes**
  * Improved semantic type inference to better detect field types from names and data types.
  * Manual semantic type selections are now preserved and no longer overwritten by automatic inference.


ChibiBlasphem (approved):
lgtm

PR #1432, opened 2026-04-17
Fix first render of the graph

william-schlegel:
fix racing condition between dargre & reactflow to display a nice graph at first render. handle resize of the window observing the container
also add a spinner while computing the graph

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved flow visualization layout and responsiveness during initialization and container resizing.
  * Added loading indicator that displays while the flow visualization is being prepared.


ChibiBlasphem (approved):
lgtm

PR #1434, opened 2026-04-17
fix old tables edition

william-schlegel:
MAR-1783: Cannot update a Person table for "no name" field

detect changes in fields after automatic semantic type inference from name and send the changes accordingly to the table update end point

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection and application of semantic type metadata changes when updating table fields. The system now properly recognizes semantic type modifications even when other field properties remain unchanged, ensuring changes are correctly persisted.


ChibiBlasphem (approved):
lgtm

PR #1440, opened 2026-04-20
show "snooze rules" button only if there are rules to snooze + displa…

Pascal-Delange:
## Summary
- Hide "Snooze Rules" button on cases with no rule hits (only shown when loaded decisions contain at least one rule with `outcome === 'hit'`)
- Fix rules section in decision panel being entirely hidden (including the "show hits only" toggle) when no rules matched — now the toggle stays visible so users can switch to viewing all rules…y fix for alerts detail panel

## Summary by CodeRabbit

* **Bug Fixes**
  * The "snooze rules" button now conditionally displays only when rule hits are detected, improving interface clarity.
  * Improved rules section rendering to properly maintain section structure while filtering execution results.


william-schlegel (commented):
lgtm

william-schlegel (approved):

PR #1441, opened 2026-04-20
Pascal/mar 1776 in cases created last 10 days by inbox the legends were all

Pascal-Delange:
## Changes
  - Fix legend showing black boxes when no data by providing explicit legend data with hardcoded colors
  - Add gradient fill (buildBarGradient), borderRadius, and borderWidth to match analytics chart style
  - Use shared nivoTheme from chart-theme.ts instead of inline theme
  - Use circle symbols in legends (symbolShape: 'circle') for consistency with analytics charts
  - Add colorBy="id" to both charts for correct color-to-key mapping

I also found that the new graph bars, while looking good in light more, are a bit violent in dark mode. Asked Maud for an opinion.


william-schlegel (approved):
lgtm

PR #1444, opened 2026-04-20
fix rule creation

OrriMandarin:
cf:
- https://linear.app/marble-eu/issue/MAR-1747/error-when-creating-a-new-past-alerts-rule
- https://linear.app/marble-eu/issue/MAR-1748/error-when-creating-a-new-screening-rule
- https://linear.app/marble-eu/issue/MAR-1749/error-when-creating-a-tag-rule

## Problem
Saving a rule of type past_alerts, screening_tags, or entity_tags failed with:

```
BadParameterError: ast node does not return a [score_computation_result]
```

## Root cause

The frontend was sending Switch.namedChildren.field = { name: "Undefined" }. Undefined is a real backend function that throws at eval time — it does not behave like a no-op. The backend's Switch evaluator expects field to be a null-valued constant when there is no field (if field != nil guard).

screening_tags rules also had a secondary bug: for every branch, the AST builder hardcoded MonitoringListCheck, even for entity_tags rules which must use RecordHasTags. The tag-check config was also missing targetTableName.

## Fix
- Switch field is now a null-valued constant node instead of Undefined. This applies to all three rule types. Serialises to { constant: null, children: [], named_children: {} }, which satisfies the backend's nil-field guard.

- entity_tags rules now build RecordHasTags nodes (not MonitoringListCheck). screening_tags still builds MonitoringListCheck. Both pass the rule's entityType as targetTableName in the config — previously sent as empty string, which would have failed backend validation.

- Default branches for new tag rules: when a user creates a new screening_tags / entity_tags rule, the UI now starts with one configurable branch + the else branch (was: only the else branch, giving the user nothing to configure).

## Refactor (small)
- transformSwitchAstNodeToModel → transformAstNodeToModel and buildSwitchAstNodeFromModel → buildAstNodeFromModel. The buildAst* function now takes a { entityType } context so tag-rule builders can set targetTableName.

- RulesTable.handleConfirm now builds the initial AST for tag and past-alerts rules via buildAstNodeFromModel, so every rule type starts from a structurally valid shape.

- Removed the defensive isSwitchAstNode(rule.ast) filter that was hiding rules with non-standard roots — all rule ASTs are Switch-rooted now; the check is redundant.

## Reviewer notes
Only past_alerts / screening_tags / entity_tags rule creation paths were changed on the wire. user_attribute / aggregate rules are structurally identical to before — their field is a real accessor so they never hit the bad Undefined path.
No backend change is needed.

## Summary by CodeRabbit

* **Refactor**
  * Scoring rule editing and viewing now behave consistently: rule conditions render and update reliably across all rule types.
  * Tag-based and past-alert conditions are handled more robustly, preventing missing/hidden conditions in the UI.
  * Simplified internal transformations reduce conditional gaps, improving stability when editing or saving rules.

### DIFF
--- a/packages/app-builder/src/components/UserScoring/RulesTable.tsx
+++ b/packages/app-builder/src/components/UserScoring/RulesTable.tsx
@@ -1,9 +1,10 @@
 import { NewAstNode } from '@app-builder/models';
 import { NewAggregatorAstNode } from '@app-builder/models/astNode/aggregation';
-import { isSwitchAstNode, NewSwitchAstNode } from '@app-builder/models/astNode/control-flow';
+import { NewSwitchAstNode } from '@app-builder/models/astNode/control-flow';
 import { type CustomList } from '@app-builder/models/custom-list';
 import { type DataModel } from '@app-builder/models/data-model';
 import {
+  buildAstNodeFromModel,
   RULE_TYPES,
   type RuleModelType,
   type ScoringRule,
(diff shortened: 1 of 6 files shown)


OrriMandarin:
I trust Claude, but I personally tested the rules.

The rules were successfully created.

I made a slight modification to the rule creation process to include two branches when creating tags and sanctions rules. Previously, only the else branch was created by default.

Pascal-Delange (approved):
Seems legit

PR #1445, opened 2026-04-20
fix flashing tab in user scoring

Pascal-Delange:
## Content
  1. Tab selection blink - After committing/updating a ruleset, the tab's link version changed (e.g. draft → 1) before the URL updated, causing aria-current to drop and the active style to flash. Fixed by matching active state on recordType
  only via data-status, bypassing Link's full-path matching.
  2. Loader data crash - router.invalidate() temporarily made useLoaderData() return undefined, crashing the child route component. Added a guard to return null during the brief undefined window.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed active state indication for ruleset tabs
  * Improved data handling during route transitions to prevent temporary display issues

### DIFF
--- a/packages/app-builder/src/components/UserScoring/ScoringSectionLayout.tsx
+++ b/packages/app-builder/src/components/UserScoring/ScoringSectionLayout.tsx
@@ -6,7 +6,7 @@ import { type UpdateScoringRulesetPayload, updateScoringRulesetPayloadSchema } f
 import { handleSubmit } from '@app-builder/utils/form';
 import { createSimpleContext } from '@marble/shared';
 import { useForm } from '@tanstack/react-form';
-import { Link, Outlet, useMatches, useNavigate, useRouter } from '@tanstack/react-router';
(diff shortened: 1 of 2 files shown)


packages/app-builder/src/components/UserScoring/ScoringSectionLayout.tsx
```diff
@@ -82,6 +70,27 @@ export function ScoringSectionLayout({ maxRiskLevel }: { maxRiskLevel: number |
   );
 }
 
+function RulesetTab({ ruleset }: { ruleset: { recordType: string; status: string; version: number; name: string } }) {
+  const params = useParams({ strict: false }) as { recordType?: string };
```
william-schlegel: unnecessary cast for params

Pascal-Delange: indeed, fixed

william-schlegel (approved):
lgtm

PR #1447, opened 2026-04-20
pass license to user scoring scope rule builder

Pascal-Delange:
Rule builder needs an optional license model input to display license restricted features (currently: std dev & percentile ops)


william-schlegel (approved):
lgtm

PR #1462, opened 2026-04-21
fix: align label and value in ingested-data grids

OrriMandarin:
## Summary
- Vertically centers the label with its value in the four grid layouts used to display ingested data.
- Fixes the visual misalignment in Customer hub, Case manager decision drawer, Screening request/query details, and Data Model Explorer.
- Single-property CSS change per file (`items-center`) — no component-behavior changes.

## Affected layouts
- `DataField` — the canonical label/value row used across Customer hub, Case manager → screening object details, Ingested object modal, Table record preview drawer, Pivot details, Decision trigger-object panel.
- `CaseDetailTriggerObject` — Case manager decision drawer, trigger-object grid.
- `DataListGrid` — shared grid used in the screening request detail panel.
- `QueryObjectDetail` inside `ScreeningQueryDetail` — screening query tab in the Case manager.

## Note on history
The branch contains a few experimental commits explored during review (baseline alignment, column-split restructure) that were reverted. The last commit (`center label and value in ingested-data grids`) represents the final shipped diff; squash-merging is recommended.

## Test plan
- [ ] Customer hub: UUID / foreign-key / GPS-coords pills line up with their labels.
- [ ] Case manager → decision drawer → trigger object: property/value rows line up.
- [ ] Case manager → screening tab → query detail: rows (type + properties) line up.
- [ ] Single-column and multi-column DataField layouts render without regressions.
- [ ] Dark mode unaffected.


Pascal-Delange (commented):
the change in the diff is not consistent with the PR desc ?

OrriMandarin:
Capture d’écran 2026-04-21 à 11 51 25
Capture d’écran 2026-04-21 à 11 56 20

The issue with this fix is that if we extend the value section (like when we want to view related ingested data), the label will be centered within this section and not align with the first element of the value.

An existing issue is that if the right section has a height element, the left section is also affected. Perhaps I can try to have independent left and right sections.

packages/app-builder/src/components/Data/DataVisualisation/DataField.tsx
```diff
@@ -599,9 +598,7 @@ function LinkToValue({ value, linkedTo }: { value?: string; linkedTo?: string })
               />
             ))
             .otherwise(() => (
-              <span className="text-grey-secondary text-s">
-                {t('data:viewer.related_object_not_ingested', { tableName: linkedTo, objectId: value })}
-              </span>
+              <EmptyValue />
```
Pascal-Delange: isn't this rolling back your previous PR ?

OrriMandarin:
Before: 
Capture d’écran 2026-04-21 à 12 10 16



After: 
Capture d’écran 2026-04-21 à 12 10 10


The only point is that the label is centered to the value part (in case we want to display the related ingested data), not aligned with the first line.

OrriMandarin:
@ChibiBlasphem Can be interesting to have your opinion on this.

packages/app-builder/src/components/Data/DataVisualisation/DataField.tsx
```diff
@@ -115,7 +115,7 @@ export function DataField({ field, value, linkedTo, metaData, currency }: DataFi
 
   return (
     <DataFieldProvider value={contextValue}>
-      <div className="col-span-2 grid grid-cols-subgrid items-start">
+      <div className="col-span-2 grid grid-cols-subgrid items-center">
```
william-schlegel, line 118: I’m not sure if vertically aligning is better than starting the alignment.
It’s usually better to have labels aligned at the top, especially when the content is tall.
Perhaps we should consult Maud about a design perspective.

packages/app-builder/src/components/Data/DataVisualisation/DataField.tsx
```diff
@@ -115,7 +115,7 @@ export function DataField({ field, value, linkedTo, metaData, currency }: DataFi
...
         <label htmlFor={field?.id} className="text-grey-secondary">
           {field?.name}
```
william-schlegel, line 120: if there are issues with very long labels, you can add a `truncate` class

PR #1465, opened 2026-04-21
fix filters in audit logs page

Pascal-Delange:
## Root cause
After migrating from Remix to TanStack Start, the audit-logs route was still using the Remix-era pattern of parsing search params from getRequest().url. TanStack Router requires explicit validateSearch on the route definition for search params to flow through.

### Changes in audit-logs.tsx:

  1. Added validateSearch: pageQueryStringSchema — tells TanStack Router to parse and validate search params from the URL
  2. Added loaderDeps: ({ search }) => search — makes the loader re-run when search params change
  3. Loader now receives search params via data input (with inputValidator) instead of manually parsing getRequest().url
  4. navigate uses search option instead of string-concatenating query params with qs — this is the TanStack Router way
  5. Switched from zod to zod/v4 to match the rest of the codebase
  6. Removed unused getRequest, QueryString imports

## Summary by CodeRabbit

* **Refactor**
  * Updated audit logs search parameter validation and pagination navigation handling for improved internal consistency.


william-schlegel (approved):
LGTM

PR #1471, opened 2026-04-21
fix immutable error message while uploading data

william-schlegel:
server fn cannot return raw fetch response, need to create a new response with new header

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved upload endpoints (ingestion, list, screening) to consistently return the upstream status, status text, and adjusted headers to clients.
  * Fixed handling of empty responses so HTTP 204/205/304 return no body, and certain transfer headers are removed to prevent client issues.


Pascal-Delange (approved):

Pascal-Delange:
LGTM, but I did not quite understand why ?

PR #1474, opened 2026-04-21
fix: keep general settings when creating new user scoring draft

Pascal-Delange:
## Summary

  - Fix general settings (scoringIntervalSeconds) being reset when adding, editing, or deleting a rule — the POST to /scoring/rulesets/{recordType} replaces the full draft, and scoringIntervalSeconds was missing from the payload in
  RulesTable.tsx
  - Fix risk level display off-by-one: the API returns 1-based risk_level but the frontend color/label arrays are 0-indexed. Applied risk_level - 1 in ClientDetailPage, ScoreDetailPanel, and ScoringOverviewPage

See also https://linear.app/marble-eu/issue/MAR-1808/after-initial-scoring-wrong-risk-level

### DIFF
--- a/packages/app-builder/src/components/ClientDetail/ClientDetailPage.tsx
+++ b/packages/app-builder/src/components/ClientDetail/ClientDetailPage.tsx
@@ -66,10 +66,12 @@ export const ClientDetailPage = ({
 
   if (scoringSettings && activeScore) {
     scoreColor =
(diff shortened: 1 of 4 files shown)


william-schlegel (approved):
The `as 3 | 4 | 5 | 6][` cast is quite prominent in my opinion, but otherwise, it functions correctly.

PR #1486, opened 2026-04-23
fix decision page query filters

Pascal-Delange:
## Fix array filters broken on decisions page (MAR-1824)

  After the Remix → TanStack Router migration, filters on scenario, trigger object, scheduled execution, and inbox all failed with ARRAY_MAX_LENGTH.

##  Root cause
Two issues compounded:

  1. adaptFilterValues spread otherFilters (from getDecisionFilters) over emptyDecisionFilters. Because getDecisionFilters explicitly sets all fields in its return object — including absent ones as undefined — the spread overrode the []
  defaults with undefined. So after any navigation that didn't include an array filter in the URL, the form would initialize those fields as undefined instead of [].
  2. With selectedValue={undefined}, SelectWithCombobox called onSelectedValueChange with a plain string (the clicked item) rather than a string[]. That string went into the form, was passed unencoded to navigate (stringifyValue passes raw
  strings through), producing ?scenarioId=<uuid> in the URL. TanStack Router's parseSearch couldn't JSON-parse a bare UUID, left it as a string, and validateSearch failed — the page never reached the loader.

  Fix: Replace the spread-based adaptFilterValues with explicit field mapping using ?? [] / ?? null defaults. Each field's default is now stated directly, making the bug class impossible and the intent obvious. emptyDecisionFilters is kept for
  the form type inference helper and useClearFilter.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved filter form handling to ensure proper normalization of empty filter selections and date range configurations.

### DIFF
--- a/packages/app-builder/src/components/Decisions/Filters/DecisionFiltersContext.tsx
+++ b/packages/app-builder/src/components/Decisions/Filters/DecisionFiltersContext.tsx
@@ -55,29 +55,28 @@ const emptyDecisionFilters: DecisionFiltersForm = {
   triggerObjectId: null,
 };
 
-function adaptFilterValues({ dateRange, ...otherFilters }: DecisionFilters): DecisionFiltersForm {
-  const adaptedFilterValues: DecisionFiltersForm = {
-    ...emptyDecisionFilters,
-    ...otherFilters,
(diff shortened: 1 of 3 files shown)


william-schlegel (approved):
LGTM

PR #1502, opened 2026-04-30
MAR-1848: Make the field details panel sticky

william-schlegel:
Enhance CreateTable and EditTable components with ScrollArea for improved scrolling experience. 
Refactor FieldsForm to support keyboard navigation for field reordering and integrate ScrollArea for better layout management.

## Summary by CodeRabbit

* **New Features**
  * Added keyboard shortcuts (arrow keys with Command/Control modifier) for reordering fields in table configuration.

* **Bug Fixes**
  * Improved scrolling behavior in table creation and editing interfaces.
  * Enhanced layout sizing for better content visibility in data forms.


Pascal-Delange (commented):
I think I'd like to leave this one for @ChibiBlasphem

packages/app-builder/src/components/Data/SemanticTables/Shared/FieldsForm.tsx
```diff
@@ -55,34 +90,49 @@ export function FieldsForm({
...
-                    </div>
-                  )}
-                </Draggable>
-              ))}
-              {dropProvided.placeholder}
-            </div>
-          )}
-        </Droppable>
-      </DragDropContext>
+      <ScrollArea.Root className="flex-1 min-h-0">
```
ChibiBlasphem: We're using ScrollArea here while other components are using ScrollAreaV2. Is it wanted?

packages/app-builder/src/components/Data/SemanticTables/Shared/FieldsForm.tsx
```diff
@@ -30,20 +31,54 @@ export function FieldsForm({
...
+            reorderFields(index, index - 1);
+            setTimeout(() => buttonRefs.current[index - 1]?.focus(), 0);
+          }
+        } else {
+          buttonRefs.current[index - 1]?.focus();
+        }
+      }
+    },
+    [fields.length, reorderFields],
+  );
```
ChibiBlasphem, line 67: The `setTimeout` here will generally work but it's not guaranteed. It would be better to set a ref and use a `useEfect` to set the focus.

packages/app-builder/src/components/Data/SemanticTables/Shared/TableForm.tsx
```diff
@@ -115,8 +115,8 @@ export function FormTable({
   );
 
   return (
-    <div className="flex gap-v2-lg">
-      <div className="flex min-w-0 flex-1 flex-col gap-v2-lg">
+    <div className="flex flex-1 min-h-0 gap-v2-lg">
+      <div className="flex flex-1 min-h-0 flex-col gap-v2-lg">
```
ChibiBlasphem: Removing the "min-w-0" here could lead to the children being able to overflow the container. I suggested restoring it

ChibiBlasphem (changes requested):

packages/app-builder/src/components/Data/SemanticTables/Shared/FieldsForm.tsx
william-schlegel: I initially used useRefs, but there were several side effects, and the focus was not always functioning correctly. I can attempt to find a more suitable approach, but the reliable setTimeout method appears to be sufficient for my needs.

packages/app-builder/src/components/Data/SemanticTables/Shared/TableForm.tsx
william-schlegel: Coderabbit seems to be fine with your idea, I will restore it

packages/app-builder/src/components/Data/SemanticTables/Shared/FieldsForm.tsx
william-schlegel: I did not see the V2 at first, but it works too (and is more simple)

ChibiBlasphem (approved):
lgtm

PR #1522, opened 2026-05-07
Mar 1858 buttons hits to review

william-schlegel:
Add more information about the matches status in the button so it is clear 

Screenshot 2026-05-07 at 16 45 12

## Summary by CodeRabbit

* **New Features**
  * Screening status badges now fetch screening details and display per-status match counts, show loading spinners and global error states, and behave interactively within alert rows.
* **Localization**
  * Added/updated translations for pending and skipped screening statuses and count placeholders in Arabic, English, and French.

[Review Change Stack](https://app.coderabbit.ai/change-stack/checkmarble/marble-frontend/pull/1522)

### DIFF
--- a/packages/app-builder/src/components/Cases/CaseAlerts.tsx
+++ b/packages/app-builder/src/components/Cases/CaseAlerts.tsx
@@ -425,11 +425,7 @@ export const ScreeningStatusBadge = ({
   return (
     <Button variant={config.variant} size="small" className="shadow-sm" tabIndex={-1}>
       {match(screeningQuery)
-        .with({ isPending: true }, () => (
-          <div className="flex items-center justify-center p-8">
-            <Spinner className="size-4" />
-          </div>
-        ))
+        .with({ isPending: true }, () => <Spinner className="size-4" />)
         .with({ isError: true }, () => (
           <div className="text-grey-secondary p-8 text-center text-s">{t('common:global_error')}</div>
         ))
--- a/packages/app-builder/src/components/Cases/CaseAlerts.tsx
+++ b/packages/app-builder/src/components/Cases/CaseAlerts.tsx
@@ -444,10 +444,7 @@ export const ScreeningStatusBadge = ({
               {Object.entries(matchCount)
                 .filter(([, count]) => count > 0)
                 .map(([status, count]) => (
-                  <div key={status}>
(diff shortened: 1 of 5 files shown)


packages/app-builder/src/components/Cases/CaseAlerts.tsx
```diff
@@ -407,9 +424,35 @@ export const ScreeningStatusBadge = ({ status, count }: { status: ScreeningStatu
...
+            R.mapValues((group) => group.length),
+          );
+          return (
+            <>
+              {Object.entries(matchCount)
+                .filter(([, count]) => count > 0)
+                .map(([status, count]) => (
+                  <div key={status}>
+                    {t(`screenings:status.${status}`)}
+                    {` (${count})`}
```
Pascal-Delange: Nitpicking about the logic in arabic: this will break RTL in arabic, the clean way would be to inject the values into the translation string and handle order in there.

Pascal-Delange: @coderabbit worth a note, if that actually works with you

william-schlegel: yes, did not think about this one

packages/app-builder/src/components/Cases/CaseAlerts.tsx
```diff
@@ -407,9 +424,32 @@ export const ScreeningStatusBadge = ({ status, count }: { status: ScreeningStatu
...
+            return <div className="text-grey-secondary p-8 text-center text-s">{t('common:global_error')}</div>;
+          }
+          const matchCount = R.pipe(
+            screeningData.matches,
+            R.groupBy((m) => m.status),
+            R.mapValues((group) => group.length),
+          );
+          return (
+            <>
+              {Object.entries(matchCount)
```
Pascal-Delange, line 444: So, just a note: this will likely not always be in the same order ? Or am I missing something ?

Pascal-Delange (approved):
Approve with a small note

packages/app-builder/src/components/Cases/CaseAlerts.tsx
william-schlegel: I think that `screeningData.matches` is always in the same shape as it is ordered so it should always be in the same order in the button

PR #1579, opened 2026-05-26
hide sanctions freshness banner on managed Marble SaaS

Djauron:
## Summary

The red *"sanctions data is not up-to-date"* banner at the top of `_app/_builder` is noise on the managed Marble SaaS instance — operators on that deployment keep the OpenSanctions dataset in sync upstream, so end users can't act on the warning. Self-hosted operators *do* run their own pipeline and need the warning, so the right move is to gate the banner by deployment, not delete it.

This PR:

1. **Reverts the full removal** (commit `7cbabe2f`) — brings back `DatasetFreshnessBanner`, its query hook, server function, repository method, model/adapter, and the three i18n keys.
2. **Adds a SaaS gate** on the server function: `getDatasetFreshnessFn` short-circuits to `{ datasetFreshnessInfo: null }` whenever `context.appConfig.isManagedMarble` is `true`. Gating server-side (rather than at the component) also skips the upstream API call entirely on SaaS — no wasted request.

The `isManagedMarble` flag is the established canonical signal in the codebase (defined in `models/app-config.ts`, populated from the backend's `is_managed_marble` DTO field, already used by the IP-whitelisting route and the sign-in pages).

## Test plan
- [x] `bun run type-check` passes across the workspace (pre-push hook)
- [x] `bun run format:check` clean
- [ ] Manual smoke (SaaS): set `is_managed_marble: true` on the backend, load any `/_app/_builder` route — banner does NOT render; Network tab should show the `opensanctions-dataset-freshness-info` query returning `datasetFreshnessInfo: null` with no upstream call
- [ ] Manual smoke (self-hosted): with `is_managed_marble: false` and the upstream dataset deliberately stale, confirm the banner DOES render at the top of the layout with the expected `lastExport` date

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed dataset freshness information handling in managed deployments by skipping unnecessary computation while preserving standard behavior for other deployment configurations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/checkmarble/marble-frontend/pull/1579?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


william-schlegel (approved):
lgtm

Pascal-Delange (changes requested):
Sorry misunderstanding 🙈 
Please only ignore the display _on our SaaS instance_

Pascal-Delange (approved):
🙏

PR #1602, opened 2026-06-04
Feat/mar 1951 scenario commit activate workflow with fewer clicks

william-schlegel:
# what has been done

## new progress/stepper component

A new ui component for progress with steps (slightly animated)

Screenshot 2026-06-04 at 11 13 10

## wizard modal for commit/prepare/activate a version

same logic than previous modals but all in one and without the checkboxes

Screenshot 2026-06-04 at 11 07 41
Screenshot 2026-06-04 at 11 14 12

## Claude rule

add a rule to avoid using deprecated components and use the default component design unless explicitly asking otherwise


ChibiBlasphem (approved):
lgtm

PR #1603, opened 2026-06-04
Feat/mar 1917 create UI to configure which screening provider to use on a feature

william-schlegel:
# New setting page for screening providers

Screenshot 2026-06-04 at 15 41 59

Based on FF lexisnexis (in organization_feature_access table) and API Key entitlement

For the moment, the configuration is for the 3 settings, It may be restricted to 2 as manual search and continuous screening should have the same value (to be confirmed)

I added a callout with the error instead of the toast because I improved error management and there are long messages returned which are difficult to read in the toast

Screenshot 2026-06-04 at 16 43 29

## Summary by CodeRabbit

* **New Features**
  * Admins can configure screening providers for manual search, transaction monitoring, and continuous monitoring via a new settings page; LexisNexis is available alongside OpenSanctions when entitled. Page includes Save/Reset controls and success/error toasts and is reachable from Settings.
* **Localization**
  * UI text for the screening providers settings added in English, French, and Arabic.


packages/app-builder/src/components/Settings/ScreeningProviders/ScreeningProvidersSettingsPage.tsx
```diff
@@ -0,0 +1,133 @@
...
+              form="screening-providers-form"
+              onClick={(e) => e.stopPropagation()}
+            >
+              {t('settings:screening_providers.save')}
+            </Button>
+          </CollapsiblePaper.Title>
+          <CollapsiblePaper.Content>
+            <Callout color="red" icon="error" iconColor="red" className="mb-v2-md">
+              {submitError}
+            </Callout>
```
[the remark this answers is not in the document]
william-schlegel: Callout renders Null when there is no children provided

packages/app-builder/src/models/feature-access.ts
```diff
@@ -52,5 +54,8 @@ export function adaptFeatureAccesses(dto: FeatureAccessDto): FeatureAccesses {
     continuousScreening: dto.continuous_screening,
     aiRuleBuilding: dto.ai_rule_building,
     userScoring: dto.user_scoring,
+    // Default to restricted while the backend rolls out the entitlement; OpenSanctions
```
Pascal-Delange: the entitlement should be there now ?

packages/app-builder/src/schemas/settings.ts
```diff
@@ -88,6 +88,16 @@ export const updateOrganizationPayloadSchema = z.object({
 });
 export type UpdateOrganizationPayload = z.infer<typeof updateOrganizationPayloadSchema>;
 
+export const screeningProviderSchema = z.enum(['opensanctions', 'lexisnexis']);
+
+export const updateScreeningProvidersPayloadSchema = z.object({
+  organizationId: z.string().min(1),
```
Pascal-Delange: Thinking out loud: in my understanding the frontend should never need to touch or even know an organizationId, because it's contained in the token.
Am I missing something ?

Pascal-Delange: (and if we'd want to type it, it would be uuid)

packages/app-builder/src/server-fns/settings.ts
```diff
@@ -334,6 +335,43 @@ export const updateOrganizationScenariosFn = createServerFn({ method: 'POST' })
     });
   });
 
+export const updateScreeningProvidersFn = createServerFn({ method: 'POST' })
+  .middleware([authMiddleware])
+  .inputValidator(updateScreeningProvidersPayloadSchema)
+  .handler(async ({ context, data }) => {
+    try {
+      await context.authInfo.organization.updateOrganization({
+        organizationId: data.organizationId,
```
Pascal-Delange, line 344: Ok I see where the orgId comes from now.
Did we enable field-aware permissions to update things on the organization on the backend?

Pascal-Delange (commented):
I confirm that manual and continuous screening should use the same provider
Otherwise, looks good to me

packages/app-builder/src/server-fns/settings.ts
ChibiBlasphem: I agree with you @Pascal-Delange, is it wanted that an endpoint from the backend needs the orgId passed as an argument?

packages/app-builder/src/models/feature-access.ts
william-schlegel: no it's because I forget to update the `feature-access-api.yaml` with lexisnexis required so it wanted a value not undefined

packages/app-builder/src/schemas/settings.ts
william-schlegel: I did not change this part, I just followed existing patterns

ChibiBlasphem (approved):
lgtm

PR #1628, opened 2026-06-10
fix MAR-1976 MAR-1981 and MAR-1992

william-schlegel:
fix minor problems:

- MAR-1992: overflow x of the page
- MAR-1981: Case status could be on the same line
- MAR-1976: Navigation option part is not align with the rest of the card

refacto `TriggerFieldsRow`in `CaseAlerts.tsx` with existing `ExpandableGroupTagLine` to avoid redundant code


ChibiBlasphem (approved):
lgtm

PR #1630, opened 2026-06-10
fix csp headers ?

Pascal-Delange:
## Restore CSP & X-Frame-Options on document responses

Our CSP and frame protection were never actually shipping — not in prod, staging, or dev. Two things broke silently during the Remix → TanStack migration:

- **Headers never reached the page.** `root.ts` set them via `setResponseHeaders` inside a `createServerFn` loader, but TanStack runs server fns in their own H3 event scope, so those headers never merge into the (2xx) streamed document.
- **Hydration scripts weren't nonced.** TanStack's `<Scripts>` reads its nonce from `router.options.ssr.nonce` (Remix's `<Scripts nonce>` equivalent), which we never set — so enforcing the CSP would block our own bootstrap scripts.

### What changed
- New `securityHeadersMiddleware` (request middleware in `start.ts`) sets CSP + `X-Frame-Options: DENY` on the real document `Response`.
- One per-request nonce now flows through a request-scoped `AsyncLocalStorage` (`utils/security-headers.server.ts`) shared by the middleware, `router.tsx` (`ssr.nonce`), and the root loader (CSP + inline `window.ENV`).
- `router.tsx` reads the nonce via `createIsomorphicFn` so `node:async_hooks` stays out of the client bundle.

### Verified
Dev + production build: headers present, nonces aligned, hydration clean (0 CSP violations), authenticated pages refuse to be framed.

⚠️ If you add inline `<script>`s, they must carry the nonce (`useNonce()` / `NonceProvider`) or they'll be blocked now that CSP is enforced.

## Summary by CodeRabbit

* **New Features**
  * Per-request CSP nonce generation with automatic CSP header injection and added X-Frame-Options for stronger protection.
  * Request-scoped storage for CSP nonce and policy to ensure consistent headers across request handling.
  * Router/server wiring to apply nonces to SSR hydration scripts.
  * Inline ENV script now suppresses hydration warnings.

* **Configuration**
  * Permissions updated to explicit permission names instead of wildcard entries.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,6 +1,6 @@
 {
   "permissions": {
-    "allow": ["Edit:*", "Write:*", "MultiEdit:*", "NotebookEdit:*", "Bash:*"],
+    "allow": ["Edit", "Write", "MultiEdit", "NotebookEdit", "Bash"],
     "defaultMode": "acceptEdits"
   },
   "hooks": {
--- a/packages/app-builder/src/middlewares/security-headers.ts
+++ b/packages/app-builder/src/middlewares/security-headers.ts
@@ -1,0 +1,30 @@
+import { getContentSecurityPolicy, runWithSecurityHeadersStore } from '@app-builder/utils/security-headers.server';
+import { createMiddleware } from '@tanstack/react-start';
+
+// Generates the per-request CSP nonce and applies security headers to every
+// document response.
+//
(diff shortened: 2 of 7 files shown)


william-schlegel (approved):
LGTM

ChibiBlasphem (approved):
lgtm

PR #1638, opened 2026-06-12
Pascal/save and read freeform search

Pascal-Delange:


william-schlegel (approved):
LGTM
So I need to call the `save` endpoint to save the search + results
Do I have all the searchs when I call the  getFreeformSearch or only the one which were saved manually?
How does the `is_saved` flag work?

Pascal-Delange:
> LGTM So I need to call the `save` endpoint to save the search + results Do I have all the searchs when I call the getFreeformSearch or only the one which were saved manually? How does the `is_saved` flag work?

By default you will get all searches when calling the list endpoint, but you can filter to get only saved ones - oh well looks like I forgot to do the last step of attaching the wires for this on the DTO in the backend, let me fix it.
is_saved is just a boolean that tells you if if results have been saved on a given search.

william-schlegel:
So maybe I can add a filter "with or without saved results" the panel if i just want to retrieve a search config to replay it?
Somewhere here
Screenshot 2026-06-12 at 13 49 12

PR #1657, opened 2026-06-18
Pascal/mar 2019 frontend pages redirect from short to long form UUID

Pascal-Delange:
## Redirect long-form UUIDs to short form

Long-form UUIDs in the URL stopped redirecting to short form after the TanStack migration (loaders threw a raw `Response(302)`, which the server-fn layer no longer honors as a redirect), so they crashed `fromSUUIDtoUUID`.

- Add a global `shortUUIDRedirectMiddleware` that 302s long-form UUID path segments to short form (skips `/client-detail`, `/ressources`, `/oidc`, and the raw trailing `$pivotValue` segment).
- Remove the now-dead `handleParseParamError` and `RawUUIDIssue` plumbing; loaders 404 on parse failure.

## Summary by CodeRabbit

* **New Features**
  * Added automatic URL redirection that converts long-form UUID path segments to canonical short-UUID form, while preserving query strings and hashes and respecting excluded route prefixes.

* **Bug Fixes**
  * Improved invalid resource ID handling by returning consistent `404 Not Found` responses when IDs can’t be parsed.
  * Simplified short-UUID validation errors to a clearer “Invalid short-uuid” message.


packages/app-builder/src/middlewares/short-uuid-redirect.ts
```diff
@@ -0,0 +1,51 @@
+import { fromUUIDtoSUUID } from '@app-builder/utils/short-uuid';
+import { createMiddleware } from '@tanstack/react-start';
+import { getRequest } from '@tanstack/react-start/server';
+
+const LONG_FORM_UUID = /^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$/;
```
william-schlegel: for consistency, as all params are parsed using zod schemas, you could do

```ts
function isLongUUID(value) {
   return z.uuid().safeParse(value).success
}
```

william-schlegel (approved):
lgtm

PR #1668, opened 2026-06-21
Pascal/session cookie

Pascal-Delange:
## fix(auth): keep users signed in — server-side token refresh + sliding session

Firebase auth stored only the short-lived Marble token server-side, so after it expired during inactivity, SSR loads bounced to sign-in (the browser's Firebase session was invisible to the server). We now persist the provider refresh token and refresh the Marble token server-side, and replace the 12h hard session cutoff with a sliding idle timeout + a longer absolute cap.

- Store the Firebase refresh token in the session at sign-in; add `refreshMarbleToken()` used both on 401 and proactively in `isAuthenticated` (also fixes the same latent inactivity-logout for OIDC)
- Idle timeout: log out after inactivity via `lastActivityAt` in the session — default 24h, `SESSION_IDLE_TIMEOUT`
- Absolute cap: cookie `maxAge` default raised 12h → 7d, `SESSION_MAX_AGE`

New env var `SESSION_IDLE_TIMEOUT` (see environment.ts). Existing sessions re-login once (no stored refresh token / activity timestamp).

Worked all good for me locally with emulator, next to test on a real firebase account.

---

I also added some tentative debug logs to try and explain a "fail to sign-in at first attempt after idle" issue I'm observing, at least in dev mode, for a long time.

## Summary by CodeRabbit

* **New Features**
  * Added support for longer-lasting sign-in sessions with refresh-based session continuity.
  * Introduced idle timeout handling to automatically end inactive sessions.

* **Bug Fixes**
  * Improved sign-in flows for email, Google, and Microsoft so authentication tokens are handled more reliably.
  * Better session recovery when access tokens expire, reducing unexpected sign-outs.
  * Updated security cookie handling to work more consistently across secure and non-secure requests.


william-schlegel (approved):
LGTM

william-schlegel (approved):

PR #1673, opened 2026-06-23
fix expandable group tag line measurement

william-schlegel:
fix measurement method to avoid layout shift

## Bug Fixes
- Modified ExpandableGroupTagLine component's pre-measurement rendering to display a conservative overflow indicator (`+N`) instead of all items, reflecting updated overflow handling logic.


ChibiBlasphem (approved):
lgtm

PR #1674, opened 2026-06-23
fix some weird edgecases in data display

william-schlegel:
## Summary by CodeRabbit

* **Style**
  * Updated the copy-to-clipboard button base styling to better handle long text with improved word-wrapping and overflow prevention.
  * Added truncation styling for string value displays in data fields.
  * Refined the data fields responsive layout so container styling applies only in 2- and 3-column modes, with improved grid sizing.
  * Adjusted the sticky screening header to remove an extra stacking layer while keeping the same spacing and layout.
  * Updated the monospace Tag appearance background to match the current surface styling.


ChibiBlasphem (approved):
lgtm

william-schlegel:
@linear MAR-2048

PR #1675, opened 2026-06-23
fix display in search select and fix the user creation form

william-schlegel:
there was an issue to revalidate the user creation form after it was on error

I updated it with `Input` component instead of `FormInput`


ChibiBlasphem (approved):
lgtm

PR #1676, opened 2026-06-23
workflow inbox creation auto-fill

Pascal-Delange:
### Before
Capture d’écran 2026-06-23 à 21 36 59

### After
Capture d’écran 2026-06-23 à 21 36 18

Plus the inbox is now auto-selected after inbox creation is done.

---
### Before
Capture d’écran 2026-06-23 à 21 47 08

### After
Capture d’écran 2026-06-23 à 21 46 45

---
## Change 3
No screenshot here, but replace useVirtualTable by useTable for table of scenarios, so that we dont' have 500ms+ rendering time after hydration and content shifting etc, when we first log in to Marble.

## Summary by CodeRabbit

* **New Features**
  * Inbox creation now supports optional post-create callbacks, custom styling, and button sizing, and the inbox selector integrates the full creation flow.
  * “Ask demo” CTA now uses a native link with safe target/rel handling when displayed.
* **Bug Fixes**
  * Inbox creation now reliably returns the created inbox ID when no redirect is used, enabling callback behavior.
* **Performance**
  * Improved TanStack Router data preloading (intent-based) for faster perceived navigation.
  * Updated detection scenarios table rendering approach.

### DIFF
--- a/packages/app-builder/src/components/Auth/SignInFirstConnection.tsx
+++ b/packages/app-builder/src/components/Auth/SignInFirstConnection.tsx
@@ -1,7 +1,6 @@
 import { Link } from '@tanstack/react-router';
 import { useTranslation } from 'react-i18next';
 import { CtaV2ClassName } from 'ui-design-system';
-import { ExternalLink } from '../ExternalLink';
 
 export function SignInFirstConnection({
   isSignInHomepage,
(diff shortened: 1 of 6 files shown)


packages/app-builder/src/components/Auth/SignInFirstConnection.tsx
```diff
@@ -18,23 +17,25 @@ export function SignInFirstConnection({
         className={CtaV2ClassName({
           variant: 'secondary',
           size: 'large',
-          className: 'w-full justify-center text-center',
+          className: 'w-full justify-center text-center h-auto min-h-10 py-sm',
```
william-schlegel, line 20: We had discussions with @ChibiBlasphem about the `2 lines buttons` and how to handle them.

The design asks for a fixed height which does not look good in that case and I initially changed the `h-10` to `min-h-10` in the Button component.

We finally came back to the fixed height of 10 with the idea that if the button is 2 lines height, the label should be shortened. Maybe we can talk about it with Maud?

Pascal-Delange: I see the idea, on the other hand this one is tricky because the button width decreases with screen width, so unless we enforce _really short_, there is not good label really.
Sure we can discuss it with Maud on Thu (but I'd still merge this PR for the time being)

william-schlegel (approved):
LGTM

PR #1678, opened 2026-06-24
fix: let user select assignee and inbox on case manager MAR-2049

ChibiBlasphem:
## Summary by CodeRabbit

* **New Features**
  * The case details page now lets users edit the assignee and inbox directly instead of only viewing them.
  * Assignee and inbox selections are now shown with interactive controls for quicker updates.


william-schlegel (approved):
LGTM

PR #1679, opened 2026-06-24
fix: scroll in case manager right panel MAR-2050

ChibiBlasphem:


william-schlegel (approved):
would be solved with you sidebar PR anymay

PR #1681, opened 2026-06-24
encode object_id containing '/' in base64

william-schlegel:
if the `obect_id` contains a `/` it is converted in base64 and decoded in the route loader

## Summary by CodeRabbit

* **New Features**
  * Improved client-detail navigation so links now safely support IDs that include special characters, including slashes, with consistent URL handling across search, hierarchy, and “see all” entry points.

* **Bug Fixes**
  * Fixed client-detail pages to properly decode encoded IDs before loading data.
  * Improved client-detail loading reliability with a stronger fallback when the primary lookup fails, helping the page resolve more consistently.


ChibiBlasphem (approved):
lgtm

PR #1682, opened 2026-06-24
bump tanstack to fix hydration issue coming from incorrect streaming

Pascal-Delange:
### Why

`/detection/scenarios` crashed during hydration in production (`Invariant failed` + React #423) once the list had ≥15 rows. Root cause was upstream in TanStack's streamed SSR: with `router.options.ssr.nonce` set, a nonce'd `<script>` was injected **mid-serialized-value**, splitting a string in the dehydration payload → `SyntaxError` → `window.$_TSR` never populated → router throws `Invariant failed` on hydrate. Threshold was just where the payload spilled into a second stream chunk; prod-only because dev doesn't stream the document the same way.

In practice, the js code contained something like this `:$R[83]=[null]},"019ef9b0-ce<script nonce="8f96db699952488dab05cd61bdd083e4">(function(t){let s;try{` with mismatched quotes.

Fixed upstream by the ScriptBuffer/streaming rewrite — TanStack/router#7362 (1.171.3) and TanStack/router#7497 "fix: fix streaming" (1.171.7), which prevents mid-tag script injection.

### What

Bump the pinned `@tanstack/*` router/start set (root `overrides` + app-builder) to the aligned latest: react-router 1.170.16, react-start 1.168.26, router-core 1.171.13, start-server-core 1.169.15, etc. Type-check + prod build green; verified `/detection/scenarios` hydrates at 15+ rows.

Sources:
- [PR #7497 — fix: fix streaming](https://github.com/TanStack/router/pull/7497)
- [PR #7362 — deferred Hydrate boundary / streaming](https://github.com/TanStack/router/pull/7362)
- [Issue #7402 (linked by #7497)](https://github.com/TanStack/router/issues/7402)
- [Issue #5511 — ssr.nonce not applied to all inline scripts](https://github.com/TanStack/router/issues/5511)
- [router-core CHANGELOG](https://github.com/TanStack/router/blob/main/packages/router-core/CHANGELOG.md)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected navigation so the Cases page opens the expected view.
  * Fixed route handling for detection scenario iteration pages, improving page resolution and access.


william-schlegel (approved):
as always, 
> the latest version fixes that

ChibiBlasphem (approved):
lgtm

PR #1684, opened 2026-06-24
fix docker build hoist error

Pascal-Delange:
Fix Docker build: ensure each workspace has a node_modules dir before the per-package COPY steps, since Bun hoisting can leave some packages (e.g. ui-icons) without a local one.

Surfaced by the tanstack bump in this PR, which rewrote bun.lock and changed what gets hoisted.


ChibiBlasphem (approved):
lgtm

PR #1688, opened 2026-06-25
fix sentry upload from frontend build + make it blocking

Pascal-Delange:
A version bump in the sentry CLI makes uploading release & source maps fail (silently), because it now depends on local CA certificates.
I also changed a config to make it blocking when this happens, because I found out totally randomly by reading the CI logs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build reliability by ensuring certificate verification works during container builds.
  * Surfaced sourcemap upload failures so release builds don’t pass when uploads fail.

* **Refactor**
  * Updated route mappings to match the current app navigation structure.

### DIFF
--- a/packages/app-builder/src/routeTree.gen.ts
+++ b/packages/app-builder/src/routeTree.gen.ts
@@ -803,7 +803,7 @@ export interface FileRoutesByFullPath {
   '/sign-in-email': typeof AppAuthSignInEmailRoute
   '/account': typeof AppBuilderAccountRoute
   '/analytics-legacy': typeof AppBuilderAnalyticsLegacyRoute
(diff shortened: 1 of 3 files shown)


packages/app-builder/src/routeTree.gen.ts
Pascal-Delange, line 806: @ChibiBlasphem not sure what's happening with this, I'm getting this diff in all of my PRs?

ChibiBlasphem: It seems that's a change related to tanstack version update but the file hasn't be regenerated

ChibiBlasphem (approved):
lgtm

PR #1699, opened 2026-07-03
williamschlegel/mar-2062-ai-summary-in-case-manager-frontend-part

william-schlegel:
# AI text display in case manager

## new AIText component

Create an AIText component to display AI results in a "token stream" way (using words as chunks instead of letters)

based on existing `useWritingText` which was improved for performance (using `requestAnimationFrame` instead of `interval` and `useLayoutEffect`)

Update components to use this one accordingly `AIReviewCard` and `AIDescription`

## display new review summary in cases

when the summary is available, display it, otherwise display the first lines of the review

## improve `AIReviewPanel`

- use the new Panel component structure
- add a "see next review" button for review navigation

## fix some bugs

- duplicate keys
- undefined values


ChibiBlasphem (approved):
lgtm

PR #1713, opened 2026-07-06
revalidate on start batch exec

Pascal-Delange:
Small UX fix on scenario home page

## Summary by CodeRabbit

* **Bug Fixes**
  * Manual scenario execution now refreshes the page data after a successful trigger, so the latest status and results appear without needing a manual reload.


william-schlegel (approved):
LGTM

PR #1714, opened 2026-07-06
add a small delay before sidebar unfolds

Pascal-Delange:
Increase delay before the sidebar expands on hover from 0ms (any time the mouse gets close to it it expands) to 200ms (still fast, but requires a level of intent)


william-schlegel (approved):
Smooth

PR #1715, opened 2026-07-06
CI allows rc versions, as backend

Pascal-Delange:
Frontend version of this backend PR https://github.com/checkmarble/marble-backend/pull/1794
Allow to create and release versions like vX.X.X-rc.Y, before we release for open-source

### DIFF
--- a/.github/workflows/build_and_deploy.yaml
+++ b/.github/workflows/build_and_deploy.yaml
@@ -5,27 +5,27 @@ on:
     inputs:
       environment:
         type: string
-        description: 'Github environment to use'
+        description: "Github environment to use"
         required: true
       version:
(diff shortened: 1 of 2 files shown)


.github/workflows/build_and_deploy.yaml
```diff
@@ -92,4 +92,4 @@ jobs:
           gcloud run deploy marble-frontend \
             --quiet \
             --region="europe-west1" \
-            --image="${{ env.IMAGE }}"
+            --image="$IMAGE"
```
Pascal-Delange, line 95: Interpolate from env rather than pass directly to shell, for hardening

apognu (approved):

PR #1723, opened 2026-07-08
move the form inside the Panel which use a portal

william-schlegel:
## Summary by CodeRabbit

* **Style**
  * Adjusted the edit settings panel layout to present the form more consistently.
  * Updated the label text sizing for “score duration” and “recalculation duration” to use the current, smaller text scale.


ChibiBlasphem (approved):
lgtm

PR #1760, opened 2026-07-20
Previous page redirect to Lose your mables page

william-schlegel:
refactor: update detection scenario data hooks in edit view components (same as decision page)


ChibiBlasphem (approved):
It works, but I think we need to further investigate why the `Route.useRouteContext()` doesn't seem to contain any data when the global `useRouteContext()` does.

william-schlegel:
from my understanding, as it is a child route, the back button remount the page without the complete context from the parent route (beforeload)

PR #1769, opened 2026-07-22
Show AI suggestion reason on screening hits

OrriMandarin:
## Summary
- Screening hits with a pending AI suggestion (e.g. "Likely false positive") now show the AI's `reason` text, between the match name/badge row and the topic tags in `MatchCard`.
- New `AiSuggestionReason` component renders a wand icon + the reason as plain static text in a purple-accented bordered box — no typewriter animation, no Markdown parsing, since `reason` is static backend data shown on every panel open (not freshly-generated content).
- Visibility mirrors the existing AI confidence badge condition (`pending` status + suggestion present), with an added non-empty-reason check; both now share one `hasAiSuggestion` boolean to avoid duplicating the condition.
- No changes to the AI suggestion data model/API/query layers — `reason` already existed end-to-end and was simply never rendered.

Design: https://www.figma.com/design/a9n93l5twcEpmthKw6cBz5/Marble---Wireframes?node-id=7489-44109&t=GoJi71AaZXBmxa60-4

## Screenshot
Capture d’écran 2026-07-22 à 15 02 53

## Test plan
- [x] `bun run type-check` clean across all packages (verified via pre-push hook)
- [x] Manual visual check in a real environment: open a screening with a `pending` hit that has an AI suggestion + reason, confirm the block renders between the name row and topic tags with correct styling, and confirm it's absent for non-pending hits, hits without a suggestion, or a suggestion with an empty reason (this could not be verified in the sandboxed dev environment — no live backend access)


docs/superpowers/plans/2026-07-22-screening-hit-ai-reason.md
```diff
@@ -0,0 +1,137 @@
+# Screening Hit AI Suggestion Reason Implementation Plan
```
william-schlegel: you should put these in your `.gitignore`

packages/app-builder/src/components/Screenings/MatchCard/AiSuggestionReason.tsx
```diff
@@ -0,0 +1,14 @@
+import { Icon } from 'ui-icons';
+
+type AiSuggestionReasonProps = {
+  reason: string;
+};
+
+export const AiSuggestionReason = ({ reason }: AiSuggestionReasonProps) => {
+  return (
+    <div className="bg-surface-card border-grey-border border-l-purple-primary text-grey-primary text-small flex items-start gap-xs rounded-sm border border-l-2 p-sm">
```
william-schlegel: There is a `<Card>` component now for the cards

docs/superpowers/plans/2026-07-22-screening-hit-ai-reason.md
OrriMandarin: Yes I will delete it, I am testing the `superpower` plugin :)

packages/app-builder/src/components/Screenings/MatchCard/AiSuggestionReason.tsx
OrriMandarin: Based on the design, I removed the Card component

william-schlegel (approved):
good for now

OrriMandarin:
See later if we need to rework the component
https://linear.app/marble-eu/issue/MAR-2146/ai-suggestion-text-component-is-not-common

PR #1773, opened 2026-07-23
fix: Ai rule generation component is displayed even if the org doesn't have access

OrriMandarin:
In the rule builder page, the AI rule builder component is not protected by the organization access feature. 
As a result, an organization without this feature will still see this component. 

## Summary by CodeRabbit

* **Bug Fixes**
  * AI-assisted rule generation is now shown only when the feature is enabled.
  * Prevented unintended formula updates and change handling when AI rule descriptions are disabled.


william-schlegel (approved):
LGTM

PR #1774, opened 2026-07-23
Add logger facility and access logs to the frontend.

apognu:
Introduces **structured logging** on the server (BFF) using [Pino]. Every incoming request is now logged automatically, and there's a simple, shared logger you can use anywhere in server-side code.

### What it does

- **Logs every request** at `INFO` with the method, path, status, latency, request ID, user agent and client IP.
- **Correlation IDs**: each request gets a `request_id`, reused from an upstream trace header (e.g. Google Cloud) when present, so logs can be tied back to a single request.
- **Two output modes**, chosen automatically:
  - **Development** — colored, easy-to-read output (a header line + the fields below it).
  - **Production** — one JSON object per line, ready to be ingested by a log platform.
- Log verbosity is configurable via the `LOG_LEVEL` environment variable (defaults to `info`).

### To do next but yolo

Actual, explicit log statements now needs to be sprinkled across the codebase where relevant. One can use the different log levels in order not to flood the output with too much details.

Import `getLogger()` from `utils/logger.server` and log wherever you need to, in any server-side code (server functions, loaders, services…):

```ts
import { getLogger } from '@app-builder/utils/logger.server';

getLogger().info({ case_id }, 'Case escalated');
getLogger().warn({ user_id }, 'Sign-in retry');
getLogger().error({ err }, 'Failed to enrich KYC');
```

The first argument is a set of fields to attach, the second is the message. Inside a request, the logger automatically includes that request's `request_id`, so related log lines stay linked together — no extra wiring needed.

@+ sous le bus.

---

TODO:

 - Add log statements throughout the codebase
 - Do we go futher and instrument this codebase through for tracing?

## Summary by CodeRabbit

* **New Features**
  * Added server-side request logging with correlation/trace IDs, latency, HTTP details, and authenticated user/organization context (excluding health checks).
  * Added `LOG_LEVEL` environment variable (validated) to control log verbosity.
  * Enhanced development logs with clearer, colorized output for structured fields.
* **Bug Fixes**
  * Kept the existing authentication/redirect/error behavior unchanged while capturing request context for consistent observability.


packages/app-builder/src/middlewares/request-logging.ts
```diff
@@ -0,0 +1,54 @@
...
+  const requestIdHeader = request.headers.get('x-request-id');
+  if (requestIdHeader) return requestIdHeader;
+
+  const traceparent = request.headers.get('traceparent');
+  if (traceparent) {
+    const traceId = traceparent.split('-')[1]?.trim();
+    if (traceId) return traceId;
+  }
+
+  return crypto.randomUUID();
```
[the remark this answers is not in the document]
apognu: Meh.

packages/app-builder/src/middlewares/request-logging.ts
```diff
@@ -0,0 +1,79 @@
...
+  return runWithLogger(child, async () => {
+    try {
+      const result = await next();
+
+      child.info(
+        {
+          ...fields,
+          status: result.response.status,
+          latency: Math.round(performance.now() - start),
+          user: getRequestUserEmail(),
```
william-schlegel: is this RGPD compliant?

apognu: Technically it is, we have a legal basis for processing with our customers and Google is one of of subprocessors. The same applies with the IP address.

That makes me think I shoudl also add the organization ID in the logs.

I'll let @Pascal-Delange say what he wants, but only storing IDs in logs is overcompliance for the sake of compliance, imho, but I'm ready to hear we still want to do that.

packages/app-builder/src/middlewares/request-logging.ts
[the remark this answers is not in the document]
Pascal-Delange: "guideline" is a strong word, you know

packages/app-builder/src/middlewares/request-logging.ts
Pascal-Delange: Perfectly fine, this is Marble user data not end user ("customer of customer") data

Pascal-Delange: agreed on org_id (and hell maybe org name for convenience if that's reasonable)

packages/app-builder/src/middlewares/request-logging.ts
```diff
@@ -0,0 +1,77 @@
...
```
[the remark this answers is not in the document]
Pascal-Delange: come on be serious

packages/app-builder/src/utils/logger.server.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  const keys = Object.keys(fields).filter((key) => !['method', 'path'].includes(key));
+
+  if (keys.length > 0) {
+    const body = keys.map((key) => `${paint('36', key)}${paint('90', '=')}${paintValue(fields[key])}`).join(' ');
+
+    entry = entry + ` ${body}`;
+  }
+
+  process.stdout.write(entry + '\n');
```
Pascal-Delange, line 71: Hum, how would the whole color thing work out on Cloud Run? I'm a bit afraid we'll get the same mess as yente logs.
Maybe activate the coloring part only when running local?

Pascal-Delange (approved):

packages/app-builder/src/utils/logger.server.ts
apognu: Pretty printing (this whole function) is already only enabled in dev mode.

On a production build, this outputs raw JSON instead.

packages/app-builder/src/middlewares/request-logging.ts
apognu: I already added the `org_id`. This is all set up from `/credentials` which does not return the organization name. And I would rather not have a separate call at that level. We can add it in the future in that DTO so it is easily available here.

PR #1782, opened 2026-07-29
feat: make match review default to 'no match' on opening popover

ChibiBlasphem:


william-schlegel (approved):
give me few hours to check this 😂

PR #1784, opened 2026-07-30
feat: add next unassigned case on screening case manager MAR-2178

ChibiBlasphem:
## Summary by CodeRabbit

* **New Features**
  * Added navigation to the next unassigned case from the continuous screening page.
  * Supports opening case links in new tabs or windows using standard modified-click behavior.


william-schlegel (approved):
that was easy...

PR #1786, opened 2026-07-30
use the same layout to display dataset updates in main page and in th…

william-schlegel:
…e panel with pagination

## Summary by CodeRabbit

* **UI Improvements**
  * Streamlined the continuous screening jobs view with a consistent six-column layout.
  * Removed the processed and total item-count column from the dataset updates panel.
  * Job durations now include the job completion time in a tooltip.
  * Improved consistency when displaying results across loaded pages.
  * The status header can display the dataset updates panel trigger when enabled.


ChibiBlasphem (approved):
lgtm

PR #1788, opened 2026-07-31
fix: remove useless button on screening entity detail panel

ChibiBlasphem:
- Change "view all" button for "eye" primary icon button to open screening entity detail panel
- Removed useless old close button in screening entity detail panel header

## Summary by CodeRabbit

* **Style**
  * Updated the entity-details action to use a primary icon-only button with an eye icon.
  * Simplified the entity details panel header by removing the close button.


william-schlegel (approved):
lgtm

PR #1797, opened 2026-08-11
feat(backoffice): bootstrap backoffice project

ChibiBlasphem:


packages/backoffice/src/components/common/GridContentLoader.tsx
```diff
@@ -0,0 +1,10 @@
+export function GridContentLoader() {
+  return (
+    <div className="col-span-full p-xl flex flex-col gap-md items-center">
+      <div
```
william-schlegel, line 4: why not `<Spinner>` component?

ChibiBlasphem: `<Spinner>` component is a component from the app-builder. Moving it right now would have make the changeset a lot bigger.

packages/backoffice/src/components/pages/licenses.tsx
```diff
@@ -0,0 +1,649 @@
...
+import { z } from 'zod/v4';
+
+/* ---------------------------------- Domain --------------------------------- */
+
+const EXPIRING_SOON_DAYS = 30;
+const DAY_MS = 24 * 60 * 60 * 1000;
+
+type EntitlementKey = keyof LicenseEntitlementsDto;
+
+const ENTITLEMENT_GROUPS: { title: string; items: { key: EntitlementKey; label: string }[] }[] = [
```
william-schlegel, line 20: All these options are hardcoded? Not from the backend?

packages/backoffice/src/schemas/features.ts
```diff
@@ -0,0 +1,19 @@
+import { z } from 'zod/v4';
+
+export const OVERRIDABLE_FEATURES = [
```
william-schlegel, line 3: these values should be in a `settings.json` separate file instead of code no? For easier maintainability I mean

packages/backoffice/src/schemas/user.ts
```diff
@@ -0,0 +1,18 @@
+import { z } from 'zod/v4';
+
+const ROLE_ADMIN = 'ADMIN';
```
william-schlegel, line 3: these values already exist elsewhere (in the backend I presume) is it good to delare them again here?

packages/backoffice/src/logo.svg
william-schlegel, line 1: necessary?

william-schlegel (approved):
LGTM

packages/backoffice/src/components/pages/licenses.tsx
ChibiBlasphem: What do you mean? EntitlementKey comes from DTO. But in case the question is does the backend send a list of existing entitlements, then the answer is "no" :)

packages/backoffice/src/schemas/features.ts
ChibiBlasphem: That could be discussed but currently that's how we manage any constant array of value throughout both apps

packages/backoffice/src/schemas/user.ts
ChibiBlasphem: I don't think the backend returns any list of the roles available, quite the same is done on the app-builder if I remember correctly

packages/backoffice/src/logo.svg
ChibiBlasphem: No, I thought I removed them all. My bad. Cleaning that

PR #1813, opened 2026-08-20
Schedule execution polling

william-schlegel:
When the user launch a batch execution on a scenario, and then go to the schedule execution page, the table shows a snapshot of the current executions and never refresh
Now, if at least one batch is running (pending or processing) the table is refreshed every second

## Summary by CodeRabbit

* **New Features**
  * Scheduled execution lists can now be filtered by scenario.
  * Scheduled execution pages refresh automatically while executions are pending or processing.
  * Pages display initial execution data immediately, then update with the latest results.

* **Bug Fixes**
  * Improved consistency when identifying executions that are still in progress.


jyrivallan:
Your second commit removes tests introduced in your first commit
They seemed relevant. May I ask what made you change your mind about them?

william-schlegel:
not really relevant as those are very basic functions, and we usually add tests only on critical libraries. Those tests were added by Claude (it likes to add tests everywhere) but I did not find them particularly necessary

jyrivallan (approved):
LGTM
Thanks for adding this, not having to refresh the page manually is really a life saver

william-schlegel:
yes, I initially did this in the continuous screening observability, then in csv ingestion, but I miss this one.

PR #1860, opened 2026-09-04
Add settings for table lifecycle management.

apognu:
To prevent table bloat and slow queries, we are adding the concept of table lifecycle to the client data model. Each table can now configure a retention period for both stale and active ingested client data.

Stale row retention deleted tombstone rows for superseded versions of an ingested object, whereas live row retention delete still active rows that were ingested before the retention period.

## Summary by CodeRabbit

* **New Features**
  * Added optional lifecycle settings when creating or editing semantic tables.
  * Configure automatic deletion periods for active rows and stale versions using months or years.
  * Lifecycle settings are saved with table configurations and can be enabled or disabled.
  * Added warnings and status indicators to clarify lifecycle behavior.

* **Bug Fixes**
  * Added validation requiring positive whole-number retention periods with clear error messages.

* **Localization**
  * Added lifecycle setting translations in English, French, and Arabic.


packages/app-builder/src/models/duration.ts
```diff
@@ -3,6 +3,58 @@ import { Temporal } from 'temporal-polyfill';
...
+      ['milliseconds', duration.milliseconds],
+      ['microseconds', duration.microseconds],
+      ['nanoseconds', duration.nanoseconds],
+    ].filter(([, amount]) => amount !== 0);
+
+    if (nonCalendarUnits.length === 0 && duration.months === 0 && duration.years !== 0) {
+      return { value: duration.years, unit: 'years' };
+    }
+    if (nonCalendarUnits.length === 0 && (duration.years !== 0 || duration.months !== 0)) {
+      return { value: duration.years * 12 + duration.months, unit: 'months' };
```
william-schlegel, line 43: you are sure of this one? years * 12?

apognu: Yeah, to get the value in months for `P2Y3M`, we do `2 * 12 + 3`. Note that it should not happen right now when only manipulating things from the frontend since we have no way to provide both years and months, we should only have `PxY` or `PxM`.

william-schlegel (approved):
LGTM

PR #1874, opened 2026-09-07
Use Rust implementation of React compiler.

apognu:


william-schlegel:
This one will need a lot of cleanup in the entire database (memo and useMemo which are used a lot)
That's why I did not activate it in the recent React 19 bump

apognu:
Yeah, and I found things that don't work as well, just there as a reminder.

PR #1894, opened 2026-09-17
invalidate server error when the rule is fixed

william-schlegel:
invalidate the server error if the formula is changed
Still possible to save a rule incomplete or with errors but if you edit one, you must change something to save it again

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation messages are now cleared when the edited formula differs from the saved formula, preventing outdated server feedback from being displayed.


ChibiBlasphem (approved):
lgtm

